PR #23, opened 2019-09-20
Use gitmodules

jpmckinney:
We can restore the shell script, but in most cases it's just one command (`git submodule update --remote --merge`) and I think it's better for users to learn a bit about git than to rely on a magic command. Note that only the URLs of private repositories (not contents) are disclosed.


odscjames:
This does have one difference to how we currently work; this would mean that the particular version of the sub-module is referenced in this repository. Previously, it was always just "the latest master" version of these private directories that was used, and I don't think that created any issues for us.

Referencing a particular version is obviously the right thing to do in some cases, but in this particular case can we briefly think about issues that would create? At the very least we would need some guidelines on when and how to make commits to this repository that updated the versions, and I would like to think to try to work out if this would create other problems.

jpmckinney:
The submodules have `branch = master` configured, and updating with `git submodule update --remote --merge` will pull the latest master. It's like in the previous setup but with fewer commands, and taking advantage of features of git instead of a more complex setup.

You can try with:

```
# Get all the repos in one command (nice!)
git clone -b gitmodules --recursive git@github.com:open-contracting/deploy.git
# Move the HEAD position of a submodule to simulate being behind origin/master
cd pillar/private
git co 9264a1b6c55422f90977297188d86d05b9c04193
cd ../..
# Update to origin/master in one plain git command (no shell script)
git submodule update --remote --merge
```

jpmckinney:
In both the past and proposed approach, users need to diligently update their local copies of the private repos before working with the repo.

With this PR, that standard practice would have to do one more thing, which is to `git commit pillar/private salt/private -m "Update gitmodules"`.

We can create a shell script (e.g. [`script/update`](https://github.com/github/scripts-to-rule-them-all)) to do the following if that would significantly ease development:

```bash
git checkout master
git pull --rebase
git submodule update --remote --merge
git commit pillar/private salt/private -m "Update gitmodules"
```

The submodules change infrequently so this won't generate many commits.

odscjames:
Oh I understand the technical stuff, it's just this is a change to our process of how we work and I just wanted to think to make sure that didn't have any knock on affects. I don't think so, tho if we start to create a branch in this repo that also needed a branch in one of the private repos it might start to get fiddly. That was probably true already tho, and I think will rarely happen.

Ok, let's give it a go. 

> We can create a shell script (e.g. script/update) to do the following if that would significantly ease development:

I wouldn't like an update to automatically commit changes to sub-modules. That should be something you do consciously by whoever edits the private repos, and it may get confusing with branches. Given it's a simple commit command I wouldn't bother making a script for it. When we expand https://github.com/open-contracting/deploy/blob/master/docs/making-changes.rst let's just make sure we doc it properly.

jpmckinney:
Sounds good!

Issue #24, opened 2019-09-20
Kingfisher - there is a failed state due to View changes


odscjames:
https://github.com/open-contracting/kingfisher-views/pull/32/files removes requirements.txt

This unfortunately means we have a failed state:

> 
> ----------
>           ID: /home/ocdskfp/ocdskingfisherviews/.ve/
>     Function: virtualenv.managed
>       Result: False
>      Comment: An exception occurred in this state: Traceback (most recent call last):
>                ..............
>               FileNotFoundError: [Errno 2] No such file or directory: '/home/ocdskfp/ocdskingfisherviews/requirements.txt'
>      Started: 09:53:07.347843
>     Duration: 369.936 ms
>      Changes:

jpmckinney:
Hmm, I didn't suspect that Process deployed the master branch of Views, since master branches in our setup are development branches…

Anyhow, the deployment can be somewhat simplified by `pip install -e git+https://github.com/open-contracting/kingfisher-views.git#egg=ocdskingfisherviews`

And then instead of changing into its directory, `ocdskingfisher-views-cli` will simply be available as a command-line tool from any directory.

odscjames:
> since master branches in our setup are development branches…

For scrape, process and views we don't have any versioning process at present tho, so it's not a "dev" branch. It's more a continuous deployment type thing - as soon as something is merged to master, it should be considered deployable. Happy to discuss that, tho a new issue would be good, let's keep this issue on track.

I'll tag @kindly to comment on method of installation, we'll chat.

odscjames:
Have chatted. Personally because you have to change into the directory to activate the virtual environment anyway it never bothered me that the command only works from one place (and I think we definitely want to keep a virtual env) - but happy to try an option that allows running from any directory.

jpmckinney:
I think if it's installed with setup.py then it will be available from any directory – once the virtualenv is activated.

Whether we install via setup.py  or requirements.txt is the discussion in https://github.com/open-contracting/lib-cove-ocds/pull/28#issuecomment-533618374

If we end up installing with requirements.txt, then we can just have the command run from one place.

odscjames:
![](https://github.trello.services/images/mini-trello-icon.png) [Make sure Python dependencies are working, including knock on effects](https://trello.com/c/0mawoVXt/116-make-sure-python-dependencies-are-working-including-knock-on-effects)

odscjames:
The original problem that prompted this ticket has been fixed and deploys are fine.

[...]

PR #25, opened 2019-09-20
Stub some updates for kingfisher-views changes

jpmckinney:
This is my first time writing Salt. We probably still want a virtualenv to pip install Views into, but not sure of the preferred virtualenv setup. Related: #24

### DIFF
--- a/salt/ocdskingfisherprocess.sls
+++ b/salt/ocdskingfisherprocess.sls
@@ -26,15 +26,17 @@ ocdskingfisherprocess-prerequisites  :
       - strace
       - redis
       - libpq-dev
+  pip.installed:
+    - name: ocdskingfisherviews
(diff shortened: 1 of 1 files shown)


jpmckinney:
I'm thinking that Views should be installed into the same virtualenv as Process. Otherwise, Views will need to install its own copy of Process (as is currently the case), which seems like it can lead to issues. It's also simpler for a user to just have to activate one virtualenv when working on the remote server, instead of having to think of which virtualenv to use.

jpmckinney:
Feel free to make additional commits to this PR, or start a new PR if this is the wrong direction.

jpmckinney:
Based on my understanding in #26, we wouldn't want Views in the prerequisites section, so closing.

jpmckinney:
Also, if Views later no longer depends on Process (https://github.com/open-contracting/kingfisher-views/issues/34), it makes less sense to follow my above suggestion of installing both in the same virtualenv.

PR #29, opened 2019-09-30
Change Kingfisher Process host to OCP domain

jpmckinney:
I'll also have to update the UptimeRobot monitor: https://uptimerobot.com/dashboard#783273034

### DIFF
--- a/pillar/ocdskingfisher_live_pillar.sls
+++ b/pillar/ocdskingfisher_live_pillar.sls
@@ -1,3 +1,3 @@
 ocdskingfisherprocess:
   web:
-    host: process.ocdskingfisher.opendataservices.coop
+    host: process.kingfisher.open-contracting.org


odscjames:
This had weird problems: https://github.com/open-contracting/kingfisher-process/issues/158

jpmckinney:
In the linked issue, only the apache config had changed (ServerAlias). Wouldn't the pillar file have to change (in any case, it couldn't hurt)?

I'm not sure that the hyphen is the issue – it seems like it's a red herring.

I notice that both `ocdskingfisherprocess.sls` and `ocdskingfisherpscrape.sls` use the same servername (`ocdskingfisher-dev`) for the `apache` macro. Wouldn't this produce two VirtualHost's with the same ServerName, which was the problem in the first StackOverflow issue?

odscjames:
> In the linked issue, only the Apache config had changed (ServerAlias). Wouldn't the pillar file also have to change as in this PR (in any case, it couldn't hurt)?

We need to do both: we need to add the ServerAlias line and make sure it works so Apache will accept the request correctly. Only then can we change the pillar variable as this pull request does. Changing the pillar request before we get Apache working properly will result in a broken system.

 > I notice that both ocdskingfisherprocess.sls and ocdskingfisherpscrape.sls use the same servername (ocdskingfisher-dev) for the apache macro. Wouldn't this produce two VirtualHost's with the same ServerName, which was the problem in the first StackOverflow issue?

I don't think that's quite the solution S.O. link had but good catch - we do. We should at least clean that up and see if that helps. I will try and do that soon.

odscjames:
[Serve API and web app from process.kingfisher.open-contracting.org](https://trello.com/c/aIgP4fhz/40-serve-api-and-web-app-from-processkingfisheropen-contractingorg)

odscjames:
> We should at least clean that up 

Done. Will  report more on https://github.com/open-contracting/kingfisher-process/issues/158

jpmckinney:
Closing in favor of open-contracting/kingfisher-process#158

PR #36, opened 2019-10-07
30 revert gitmodules

odscjames:


odscjames:
Mindful of your comment about "magic scripts", I also quickly rewrote that section of docs to make clear what it's trying to achieve, and that people can use other means of achieving that if they want.

jpmckinney:
Shall we use the pattern of `script/update` from https://github.com/github/scripts-to-rule-them-all ? If we're going to have such scripts, I prefer to have them in predictable locations.

PR #44, opened 2019-10-22
Fix robots.txt for OCDS documentation

jpmckinney:
robots.txt is perhaps too clever by half. For two years we haven't been indexing OCDS for PPPs in Google… Anyhow, this should fix it.


odscjames:
And deployed

jpmckinney:
Noting that the careful restrictions around `latest` were carried over from when we wanted to prevent access to languages that had not yet been translated (this no longer occurs): https://github.com/OpenDataServices/opendataservices-deploy/commit/f223c04fc49971b5e6771c620cd4c693b405fbe2#diff-411831a7ddd321618ddf85ab69d0c291

PR #47, opened 2019-10-24
Add docs tests and fix/change Apache config

jpmckinney:
This fixes and improves the Apache config for the OCDS documentation live server.

Fixed:

- Use OCDS for PPPs versions, not OCDS versions, for OCDS for PPPs language switcher
- Add '1.0' version for OCDS for PPPs, instead of proxying to staging server
- Add 'it' to OCDS languages and configure its 404, instead of using default Apache 404 page
- Add 'en' to OC4IDS languages and configure its 404, instead of using default Apache 404 page
- Fix /infrastructure/ redirection to add version, instead of language, if version is missing
- Add language (if missing) for development branches of OC4IDS, instead of responding with directory listing
- Redirect extension pages for 1.0 branch of OCDS for PPPs, not only latest branch

Improved:

- Add tests for Apache configuration
- Avoid an extra redirect due to a missing slash when adding version
- Avoid an extra redirect due to a missing slash when adding language
- Avoid an extra redirect by using HTTPS for www.open-contracting.org
- Don't hardcode OCDS for PPPs languages anywhere
- Use string instead of float for OCDS version number 1.0
- Remove non-existing files from RedirectMatch pattern in Directory directive (index.html, switcher)
- Use https instead of https? in RewriteCond, since SSL is forced now
- Improve version switcher to be aware of inter-compatible sitemaps
- Use loops instead of copy-pasting

Changed:

- Set versions and languages for OCDS and profiles in one variable

----

If you checkout 2c00be4, you can deploy the new configuration to only the testing server. You can then checkout the last commmit and run `env FQDN=testing.live.standard.open-contracting.org pytest` to see if all tests pass (not all the tests in 2c00be4 pass presently for `standard.open-contracting.org`, as it lacks above fixes).

I unfortunately can't deploy because Salt is broken on macOS Catalina (for now).


jpmckinney:
[Update Apache config for OCDS docs](https://trello.com/c/HoeIyDJQ/137-update-apache-config-for-ocds-docs)

jpmckinney:
If this is merged, I can update ocds-docs-staging.conf and ocds-docs-live.conf to include any common configuration (instead of copy-pasting between the two, which will just lead to more work and headaches later).

PR #48, opened 2019-10-24
Small improvements to Apache config for OCDS docs

### LINKED ISSUE #2: Comment out HTTP/HTTPS Fix in Docs to Cove proxy
https://github.com/open-contracting/deploy/blob/master/salt/apache/ocds-docs-live.conf.include#L249

```
    {# This solves a problem - "Convert to Spreadsheet" was working on HTTPS but not HTTP. Fix this by forcing header to be something the CSFR check likes. #}
    {# When we change HTTPS to force, there will be no HTTP traffic and this block should be removed. #}
    {% if testing  %}
        <Location /review>
                Header add referer "https://testing.live.standard.open-contracting.org"
                RequestHeader set referer "https://testing.live.standard.open-contracting.org"
        </Location>
        <Location /infrastructure/review>
                Header add referer "https://testing.live.standard.open-contracting.org"
                RequestHeader set referer "https://testing.live.standard.open-contracting.org"
        </Location>
    {% endif %}
    {% if not testing %}
        <Location /review>
                Header add referer "https://standard.open-contracting.org"
                RequestHeader set referer "https://standard.open-contracting.org"
        </Location>
        <Location /infrastructure/review>
                Header add referer "https://standard.open-contracting.org"
                RequestHeader set referer "https://standard.open-contracting.org"
        </Location>
    {% endif %}
```

jpmckinney:
Builds on #47, Closes #2

Make Apache config for OCDS docs easier to read.

- Collect unique statements near top of file.
- Put banner statements together.

Also remove:

- ProxyPass for /i18n, which as far as I can tell is not / no longer returned by the Data Review Tool.
- legacy Directory directive, which already inherits its contents from the other Directory directive.
- review Location directive, as HTTPS is now forced and it can be removed per comments.

----

I'm not sure if order of precendence affects any of the above (I think not?).

### DIFF
--- a/salt/apache/ocds-docs-live.conf.include
+++ b/salt/apache/ocds-docs-live.conf.include
@@ -6,7 +6,7 @@
             '/profiles/ppp': {
                 'versions': ['latest', '1.0'],
                 'languages': ['en', 'es'],
-                'excludes': ['extension', 'schema'],
+                'excludes': ['schema', 'extension'],
             },
             '/infrastructure': {
                 'versions': ['latest', '0.9'],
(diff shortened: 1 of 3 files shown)


jpmckinney:
Oops, I deployed this while doing the other commits for staging. Doesn't seem to have broken anything based on tests, and master should reflect what has been deployed, so merging.

PR #49, opened 2019-10-25
Change servername and allowedhosts for OC4IDS Data Review Tool.

### LINKED ISSUE #17: Change URL used for OC4IDS Data Tool
Currently the standard site proxies traffic to the data tool via the FQDN "oc4ids.cove.live.opendataservices.coop"

https://github.com/open-contracting/deploy/blob/master/pillar/live_pillar.sls#L3

We should just change that to "cove.cove-live.oc4ids.opencontracting.uk0.bigv.io", like the other server. That will need to be changed on the Data Tool server too, and make sure the SSL cert is obtained correctly.

jpmckinney:
- Move variables that change between CoVEs to top of pillar files, to make them easier to compare
- Set variables in pillar files instead of having default values in cove.sls
- Remove default values from cove.sls
- Remove IATI logic

Closes #17. There might be more to do (not sure how to address the SSL comment in #17).

### DIFF
--- a/pillar/cove_oc4ids_live_pillar.sls
+++ b/pillar/cove_oc4ids_live_pillar.sls
@@ -1,6 +1,9 @@
 # Values used only on the Cove OC4IDS server
 default_branch: 'master'
 cove:
+  servername: 'cove-live.oc4ids.opencontracting.uk0.bigv.io'
+  allowedhosts: '.oc4ids.opencontracting.uk0.bigv.io,.standard.open-contracting.org'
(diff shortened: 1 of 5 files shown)


jpmckinney:
I had to manually delete the old certs in `/etc/letsencrypt/live/oc4ids.cove.live.opendataservices.coop` but otherwise this just worked, minus a hiccup described in #58. Still up:

https://standard.open-contracting.org/infrastructure/review/
https://cove-live.oc4ids.opencontracting.uk0.bigv.io/infrastructure/review/

Issue #35, opened 2019-10-02
Replace occurrences of old names of products, when deploying to new server


jpmckinney:
'ocdskit-web' will become confusing over time, as we (and new users) will forget that it was the old name for Toucan.

odscjames:
Is that https://github.com/open-contracting/deploy/blob/master/salt-config/roster#L13 ?

In fairness we could probably rename most of those machine names :-) They are only used internally by salt. Need to be careful tho, to make sure right variables are still applied.

jpmckinney:
'ocdskit-web' also occurs in filenames and in other files. It'll be less confusing in future if we use Toucan consistently across all repos, documentation, etc.

The other machine names at least roughly correspond to FQDNs, so they are not going to be confusing in the same way.

odscjames:
Fair point. When doing this we'll have to be a bit careful; if you just change the salt rule to install a Apache config called "toucan" instead of "webkit", say, it won't automatically remove the old one - so you'll end up with 2 Apache configs, and so on. Might be a bit of downtime, just 5 mins or so. All doable tho; prioritise away.

odscjames:
![](https://github.trello.services/images/mini-trello-icon.png) [Toucan: Rename some ocdskit-web occurrences  in deploy repository](https://trello.com/c/MZSCy5kj/123-toucan-rename-some-ocdskit-web-occurrences-in-deploy-repository)

jpmckinney:
As part of this we can start documenting what types of Salt configuration changes require removing earlier configurations (e.g. renaming configuration files).

jpmckinney:
There are now only 3 occurrences in `toucan_pillar.sls` and `toucan.sls`. Changing those are lower priority.

jpmckinney:
Similarly, there are a few occurrences of `scrape` instead of `collect` viz. Kingfisher Collect.

jpmckinney:
Toucan will be decommissioned, so that is no longer an issue.

Once the new flattening tool is deployed, Toucan will just be a web interface to OCDS Kit. However, web analytics show very low usage (low tens of users per month, mostly from Paraguay, the UK and Canada, which is where our teams work). To free up time and resources for more impactful work, Toucan will therefore be decommissioned. People who can’t install OCDS Kit locally will be referred to the OCDS Kit tutorial notebook, and Flatten Tool users will be directed to the new flattening tool.

jpmckinney:
[...]

PR #54, opened 2019-10-28
Rename ocdskit-web to toucan where this would have no impact on deployments

jpmckinney:
I did the easy renaming related to #35. We can leave that issue open for the renames that involve  moving / deleting old directories, config files, users, etc.


jpmckinney:
Merging as this is going to start causing conflicts with new PRs.

PR #55, opened 2019-10-28
Cleanup: banner_message is not used

jpmckinney:
As I work on files, it's less work (for me and any person new to the project) if I don't have to think about variables and logic that are not used.

### DIFF
--- a/salt/apache.sls
+++ b/salt/apache.sls
@@ -20,11 +20,6 @@ apache2:
     - require:
       - pkg: apache2
 
-{% if 'banner_message' in pillar %}
-substitute:
-  apache_module.enabled
-{% endif %}
-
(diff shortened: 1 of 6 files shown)


jpmckinney:
In other contexts, I've typically set an environment variable and allowed each service to decide how to present itself in different contexts (like we do with ocds-docs-*).

I can see the utility of a generic solution, but right now we don't foresee dev or tmp servers. We can consider the options if/when that becomes the case, but for now I think it's fine to remove this.

How does that sound?

jpmckinney:
Merging recent PRs as @odscjames will be busy next week, and conflicts will be hard to avoid otherwise. This one is fairly uncontroversial as it deals with code whose use is not foreseeable.

Issue #19, opened 2019-08-29
Invalid HTTP_HOST header


jpmckinney:
Invalid HTTP_HOST header: '46.43.2.235'. You may need to add '46.43.2.235' to ALLOWED_HOSTS.
Invalid HTTP_HOST header: '46.43.2.235:443'. You may need to add '46.43.2.235' to ALLOWED_HOSTS.
Invalid HTTP_HOST header: 'live.standard-search.opencontracting.uk0.bigv.io'. You may need to add 'live.standard-search.opencontracting.uk0.bigv.io' to ALLOWED_HOSTS.

Let's add these to the allowed hosts (the IP is for the standard-search server).

odscjames:
> Invalid HTTP_HOST header: '46.43.2.235'. You may need to add '46.43.2.235' to ALLOWED_HOSTS.
> Invalid HTTP_HOST header: '46.43.2.235:443'. You may need to add '46.43.2.235' to ALLOWED_HOSTS.

Is there a use case for this one? Currently I think the only reason we would see this is bots trying to crawl us. (In fact, on a lot of our servers we set up an default apache placeholder site for bots to see).

> Invalid HTTP_HOST header: 'live.standard-search.opencontracting.uk0.bigv.io'. You may need to add 'live.standard-search.opencontracting.uk0.bigv.io' to ALLOWED_HOSTS.

www.live.standard-search.opencontracting.uk0.bigv.io is however already set up as an ALLOWED_HOSTS and an SSL cert name - for the same avoiding-bots reason as above, can we leave it as this?

jpmckinney:
Aha. Is there a way to quiet specific HTTP_HOST errors, so that we don't continually get new reports in Sentry?

odscjames:
I see the problem - we do have an default host in place for HTTP crawlers, but not for HTTPS crawlers - ie h t t p://46.43.2.235/ gets placeholder, h  t t p s://46.43.2.235/ gets site and generates a sentry error.

We can look into that.

odscjames:
![](https://github.trello.services/images/mini-trello-icon.png) [Look into stopping HTTPS crawlers hitting real sites and generating errors](https://trello.com/c/tdbyqzFV/93-look-into-stopping-https-crawlers-hitting-real-sites-and-generating-errors)

jpmckinney:
This has been solved in a few different ways:

[...]

PR #56, opened 2019-10-28
Add default SSL page

jpmckinney:
Related: #19

I copied default-ssl.conf from ocds-docs-live. See also: https://gist.github.com/devlim/5b7e09ba17f2ee2abcfcd80d69cb19eb

### DIFF
--- a/salt/apache.sls
+++ b/salt/apache.sls
@@ -39,3 +39,4 @@ substitute:
 
 # Ensure 000-default conf exists, so it's obvious when we've typo'd something
 {{ apache('000-default.conf') }}
+{{ apache('default-ssl.conf') }}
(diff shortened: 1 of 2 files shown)


jpmckinney:
I'm not sure that this is enough; I think this will result in an insecure connection browser warning.

odscjames:
Having more than one SSL certificate on a ip/port combination gets difficult. Your user agent has to support SNI. While that's probably getting more and more of a safe bet these days, HTTP2 shakes this up I think - tho I need to read up on my HTTP2. But I would want to check to make sure any future move to HTTP2 won't cause issues - we are going to have to do that move some day.

jpmckinney:
At minimum I would need to set the `https` parameter to the apache macro. I started looking at https://www.digicert.com/ssl-support/apache-multiple-ssl-certificates-using-sni.htm and figured this would likely involve some changes to how `_common.conf` is used.

PR #60, opened 2019-10-29
Cleanup unused files and variables, especially for Apache

jpmckinney:
Getting a bit lost with files and variables that are unused and conditions that are never met, so cleaning up.

Add `?w=1` to the "Files changed" tab's URL to ignore whitespace changes: https://github.com/open-contracting/deploy/pull/60/files?w=1


jpmckinney:
Okay, I think this is enough cleanup for one PR. In short:

* For HTTPS websites, salt/apache had a lot of *.conf files *which are never deployed* because `_common.conf` is used instead
* For HTTP websites, salt/apache had a lot of *.conf.include files *which are never included*, except for one, which I just merged into its `.conf` file
* There were macros which were used only once. By getting rid of them, you can now read the file from top to bottom, instead of going to the bottom to read parameter values, then start from the top again, trying to reconcile what the values of macro-internal variables are…

jpmckinney:
Note: After merge, need to run:

```
salt '*' file.remove /etc/apache2/sites-available/000-default.conf.include
salt '*' file.remove /etc/apache2/sites-available/prometheus-client.conf.include
salt 'cove*live*' file.remove /etc/apache2/sites-available/cove.conf.include
salt 'ocds-docs-live' file.remove /etc/apache2/sites-available/ocds-legacy.conf.include
salt 'ocdskingfisher-new' file.remove /etc/apache2/sites-available/ocdskingfisherprocess.conf.include 
salt 'ocdskingfisher-new' file.remove /etc/apache2/sites-available/ocdskingfisherscrape.conf.include 
salt 'toucan' file.remove /etc/apache2/sites-available/ocdskit-web-redirects.conf.include
```

jpmckinney:
Merging recent PRs as @odscjames will be busy next week, and conflicts will be hard to avoid otherwise. This one shouldn't be controversial, per my description in previous comment.

jpmckinney:
And deployed (plus above commands).

odscjames:
Quick comment, 

> There were macros which were used only once.

They are only being used once at the moment, but that doesn't mean to say they will always be. 

Specifically, the reason we have such Macros are for when we have dev servers where we want to have more that one install of the app on; for example we have a Cove dev server that is capable of hosting multiple deployed branches of Cove at once on different subdomains. It is called here: https://github.com/OpenDataServices/opendataservices-deploy/blob/master/salt/cove.sls#L207

Without the macros, if variables are used fully it's possible to have a dev server with a different dev install of the app on, but you can only have app install per server.

jpmckinney:
I'll create a new issue with an alternative to using macros. Getting rid of the macros allowed me to reorganize all Django apps to use common IDs and templates. Now that this has been simplified, it will be easier to add support for multiple apps on one server.

Issue #13, opened 2019-08-21
Server monitoring


odscjames:
https://ocdsdeploy.readthedocs.io/en/latest/server-monitoring.html

Document how to fully set up the agent on each server so that others can do this. Document how to set up a server, or more likely, link to existing documentation.

odscjames:
![](https://github.trello.services/images/mini-trello-icon.png) [Docs - Server Monitoring](https://trello.com/c/1fP0KA5X/75-docs-server-monitoring)

jpmckinney:
This is now using Prometheus, whose configuration is in opendataservices-deploy. Once we setup an OCP server, we can document this further.

odscjames:
In order that one Trello card currently in progress is not blocked by another one that has been not prioritised yet, can we make this issue just about documenting client setup and add to #31 that #31  should also include adding server documentation?

jpmckinney:
@odscjames Yes, sounds good to me.

PR #85, opened 2019-11-14
Docs - Monitoring

odscjames:
https://github.com/open-contracting/deploy/issues/13

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -6,9 +6,28 @@ This section of the documentation describes facts about our servers and deployme
 Monitoring
 ----------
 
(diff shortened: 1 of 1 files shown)


odscjames:
[Deploy: Docs about monitoring servers](https://trello.com/c/1fP0KA5X/75-deploy-docs-about-monitoring-servers)

jpmckinney:
I don't see where we use the black box exporter?

odscjames:
This is a service that runs on the ODS server, so there is nothing for it in this repository at the moment. When we set up an OCP Prometheus server that will be part of it, and it will be added to this repository then.

PR #89, opened 2019-12-05
Add new kf process server

robredpath:
@jpmckinney this refactors the kf-process stuff a little and adds the new server. We can remove the old server once we decommission it.


jpmckinney:
Please create an issue to decommission the old server.

I figure we also want an issue to add `ocdskingfisherarchiveonprocess` back to the new server.

jpmckinney:
I created the issues. Merging.

Issue #31, opened 2019-10-01
Set up OCP Prometheus server


odscjames:
From #28 

Question 1: Which box should this be on - a new box? A 1GB Bytemark box should be fine.

Question 2: For alerts, we need a way to send email. What shall we use? AWS has an email sending service, or are there other options?

jpmckinney:
I'm fine with a new box.

I can create an AWS SES service. We already using AWS, so simpler than creating a new account with SendGrid, Mailgun, Mandrill, etc., etc.

jpmckinney:
Alternately, if postfix and/or sendmail work, we can just use those.

odscjames:
We have used Postfix before, and it will work IF we get all the anti-spam etc stuff right and depending on which address we are sending to. If your happy with AWS SES that might be more reliable, and given the nature of that that's an important consideration.

I'll make Trello card to allow prioritisation.

odscjames:
![](https://github.trello.services/images/mini-trello-icon.png) [Set up OCP Prometheus server](https://trello.com/c/OcivonLE/118-set-up-ocp-prometheus-server)

jpmckinney:
Sounds good – let me know when ready and I'll set up the service.

odscjames:
Note this task should also include adding documentation on the server.

odscjames:
This has started. New server created at Bytemark.

@jpmckinney  Do you want to set up 2 subdomain aliases for us to use, something like:

* server.prometheus.open-contracting.org
* alerts.prometheus.open-contracting.org

Can you set us up the email service and send on details?

Thanks

ToDo:

* Set up SSL
* Document in docs
* Create public templates of new private files
* Set up email service
* Test an alert

jpmckinney:
Why do we need two subdomains (instead of just e.g. `prometheus.open-contracting.org`)?

jpmckinney:
I added a CNAME from prometheus.open-contracting.org to prometheus.default.opencontracting.uk0.bigv.io, but I'm happy to substitute for the other hostnames once I understand why we'd want to have multiple.

jpmckinney:
I'll need to know which hostname to use to setup SES (DKIM, etc.).

odscjames:
> Why do we need two subdomains (instead of just e.g. prometheus.open-contracting.org)?

There are 2 separate services and thus 2 separate websites. The alerting stuff is actually a separate software component in Prometheus - see separate download in https://prometheus.io/download/ 

[...]

PR #102, opened 2020-02-11
Prometheus - Docs and templates for private files

odscjames:
https://github.com/open-contracting/deploy/issues/31

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -43,10 +43,12 @@ In the alerting service you can also put in a temporary "silence", with definiti
 To add a new server to Prometheus:
 
 -  Make sure ``prometheus-client-apache`` or ``prometheus-client-nginx`` is included for the server in ``salt/top.sls``.
--  Salt will try to work out a domain name to use for the client server automatically. If it gets this wrong, you can set one manually by setting the ``prometheus.client_fqdn`` variable. Make sure you set this variable for one server only, and not all servers!
--  Deploy to the server you want to moniter. This should now be working - test this by going to the end point in a web browser.
--  Now update ``salt/private/prometheus-server-server/conf-prometheus.yml`` and add details of the new end point.
--  Deploy to the Prometheus server. While you do so, watch the Status / Targets page in the monitoring service and make sure the server can access the new end point without problems.
+-  Deploy to the server you want to monitor.
+-  For Apache, Salt will try to work out a domain name to use for the client server automatically. It might fail to do this, if the server is not aware of it's own domain name (Bytemark servers tend to be fine, Hetzner tend not to). Check the host name in ``/etc/apache2/sites-enabled/prometheus-client.conf``. If it gets this wrong, you can set one manually by setting the ``prometheus.client_fqdn`` variable (You can also change the ``prometheus.client_port`` variable). Make sure you set these variables for one server only, and not all servers! Deploy again.
+-  For Nginx, it will by default serve the documents on a different port, as defined in ``salt/nginx/prometheus-client`` (currently 9158).
+-  Test this by going to the endpoint in a web browser. You should know the URL from the steps above. The username is ``prom`` and the password is in the ``prometheus.client_password`` variable. Click ``Metrics`` and you should see a text response, listing variables and values.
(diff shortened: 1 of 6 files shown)


odscjames:
[Set up OCP Prometheus server](https://trello.com/c/OcivonLE/118-set-up-ocp-prometheus-server)

docs/reference/index.rst
```diff
@@ -32,9 +32,22 @@ For access details, check the configuration file ``pillar/private/prometheus_pil
...
-Currently, Open Data Services runs a Prometheus server to process client data, which raises alarms to ODS staff only (`#31 <https://github.com/open-contracting/deploy/issues/31>`__).
+The alerting service handles actually sending alarms. Alarms are raised to relevant ODS staff and OCP staff.
+
+In the alerting service you can also put in a temporary "silence", with definitions of what alarms will be silenced. This can be used to avoid alerts if you know you are about to do something that would generate an alarm, such as taking a server off-line.
+
+To add a new server to Prometheus:
+
+-  Make sure ``prometheus-client-apache`` or ``prometheus-client-nginx`` is included for the server in ``salt/top.sls``.
+-  Salt will try to work out a domain name to use for the client server automatically. If it gets this wrong, you can set one manually by setting the ``prometheus.client_fqdn`` variable. Make sure you set this variable for one server only, and not all servers!
+-  Deploy to the server you want to moniter. This should now be working - test this by going to the end point in a web browser.
```
jpmckinney: Typos: moniter -> monitor, end point -> endpoint (again below)

jpmckinney: What is the endpoint? Can you add a link or a sample URL?

docs/reference/index.rst
```diff
@@ -32,9 +32,22 @@ For access details, check the configuration file ``pillar/private/prometheus_pil
...
 
-Currently, Open Data Services runs a Prometheus server to process client data, which raises alarms to ODS staff only (`#31 <https://github.com/open-contracting/deploy/issues/31>`__).
+The alerting service handles actually sending alarms. Alarms are raised to relevant ODS staff and OCP staff.
+
+In the alerting service you can also put in a temporary "silence", with definitions of what alarms will be silenced. This can be used to avoid alerts if you know you are about to do something that would generate an alarm, such as taking a server off-line.
+
+To add a new server to Prometheus:
+
+-  Make sure ``prometheus-client-apache`` or ``prometheus-client-nginx`` is included for the server in ``salt/top.sls``.
+-  Salt will try to work out a domain name to use for the client server automatically. If it gets this wrong, you can set one manually by setting the ``prometheus.client_fqdn`` variable. Make sure you set this variable for one server only, and not all servers!
```
jpmckinney: In what cases does it get it wrong? Does it currently get it wrong? How can we tell if it's wrong?

pillar/private-templates/prometheus_pillar.sls
```diff
@@ -3,4 +3,15 @@ prometheus:
   # use "openssl passwd -apr1" to generate this. Use the same password as above.
   client_password_as_nginx_file: "prom:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
   client_fqdn:
+  client_port: 80
```
jpmckinney: Aren't some of these not private? Same question about others.

odscjames:
> Aren't some of these not private? Same question about others.

Yeah, technically a lot of what is there isn't secret like passwords. But in our own Prometheus installation, we raised a discussion about the fact that having people knowing the exact limits and things we are monitoring might be bad if people who are deliberately trying to mess with our servers can use that knowledge to avoid tripping our alarms. As the grumpy people say, "Obscurity is not security" is an incomplete saying, it's really "Obscurity is not security if that is the ONLY security measure you have". Sometimes alongside security a bit of obscurity can be good too. Anyway, we decided to just keep a lot of our Prom setup private until we could decide how to settle this. 

But this pull request just puts in place docs that reflect what we currently have - can we put this in now, and have a separate issue to discuss making some of the currently private setup public, if that's something you want to do?

I'll do the other points now.

odscjames:
Updated

odscjames:
@jpmckinney Are you happy with this update? Do merge if so.

PR #129, opened 2020-04-16
docs: Add some notes on Maintain a server

odscjames:


odscjames:
["being responsible for OCP servers" page in docs](https://trello.com/c/vAbqECmA/194-being-responsible-for-ocp-servers-page-in-docs)

jpmckinney:
This page is designed as short [how-to guides](https://documentation.divio.com/how-to-guides/) for specific tasks.

For the "being responsible" card, I think Rob and I were thinking of a new page that outlines what it means to be responsible in a more explanatory way. It would like to specific tasks where relevant.

jpmckinney:
Rather than going back and forth, I simply wrote the content I was expecting.

Issue #130, opened 2020-04-22
Current cove-ocds master does not deploy due to salt's handling of requirements files


Bjwebb:
I just updated requirements in cove-ocds with the latest version of pip-tools https://github.com/open-contracting/cove-ocds/pull/32
This adds a bunch of comments along the lines of `# via -r requirements.in, libcove, libcoveocds, libcoveweb`.
Salt does its own processing of requirements includes, which doesn't respect comments. With the comment above, this tries to fetch a file called `requirements.in, libcove, libcoveocds, libcoveweb`, and fails.
This causes the deploy to fail:

```
         ID: /home/cove/cove/.ve/
    Function: virtualenv.managed
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/state.py", line 1981, in call
                  **cdata['kwargs'])
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/loader.py", line 1977, in wrapper
                  return f(*args, **kwargs)
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/states/virtualenv_mod.py", line 320, in managed
                  **kwargs
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/modules/pip.py", line 680, in install
                  user=user
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/modules/pip.py", line 351, in _process_requirements
                  reqs = _resolve_requirements_chain(requirement)
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/modules/pip.py", line 281, in _resolve_requirements_chain
                  chain.extend(_resolve_requirements_chain(_find_req(req_file)))
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/modules/pip.py", line 281, in _resolve_requirements_chain
                  chain.extend(_resolve_requirements_chain(_find_req(req_file)))
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/modules/pip.py", line 257, in _find_req
                  with salt.utils.files.fopen(link) as fh_link:
                File "/var/tmp/.root_ddb04c_salt/pyall/salt/utils/files.py", line 399, in fopen
                  f_handle = open(*args, **kwargs)  # pylint: disable=resource-leakage
              IOError: [Errno 2] No such file or directory: u'requirements.in, libcove, libcoveocds, libcoveweb'
```

[...]

PR #131, opened 2020-04-22
Run pip directly

Bjwebb:
This is required to avoid issues with Salt's virtualenv state, and '-r' inside comments added by pip-compile of pip-tools. https://github.com/open-contracting/deploy/issues/130

### DIFF
--- a/salt/django.sls
+++ b/salt/django.sls
@@ -77,7 +77,6 @@ django-deps:
   virtualenv.managed:
     - python: /usr/bin/python3
     - user: {{ pillar.user }}
-    - requirements: {{ djangodir }}requirements.txt
     - require:
       - pkg: django-deps
       - git: {{ pillar.git.url }}{{ djangodir }}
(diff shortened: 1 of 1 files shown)


salt/django.sls
```diff
@@ -77,21 +77,28 @@ django-deps:
...
     - user: {{ pillar.user }}
-    - requirements: {{ djangodir }}requirements.txt
     - require:
       - pkg: django-deps
       - git: {{ pillar.git.url }}{{ djangodir }}
       - file: set_lc_all # required to avoid unicode errors for the "schema" library
     - watch_in:
       - service: apache2
 
+pip_install_requirements:
```
jpmckinney, line 87: Please add a comment:

```python
# Due to a bug in Salt, we can't use `- requirements: {{ djangodir }}requirements.txt` in the above
# `{{ djangodir }}.ve/` state. See https://github.com/saltstack/salt/issues/56514
```

That way, once the bug is fixed, we can revert to a simpler Salt file.

Bjwebb:
Thanks, have added.

jpmckinney:
Merging to deploy Toucan.

Issue #144, opened 2020-05-11
kingfisher-process: Changes to postgresql.conf are not in Salt


jpmckinney:
An easy way to see non-default values:

```sql
SELECT name, current_setting(name), SOURCE
  FROM pg_settings
  WHERE SOURCE NOT IN ('default', 'override');
```

https://wiki.postgresql.org/wiki/Server_Configuration

jpmckinney:
![](https://github.trello.services/images/mini-trello-icon.png) [Kingfisher: Changes to postgresql.conf are not in Salt](https://trello.com/c/ssC6B4tA/222-kingfisher-changes-to-postgresqlconf-are-not-in-salt)

jpmckinney:
Once this is done we can create a new issue to tune PostgreSQL: https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server

I see `shared_buffers` is already changed, for instance.

odscjames:
I installed an Ubuntu 18 LTS and postgres-11 from the same package. I then compared /etc/postgresql/11/main. pg_hba.conf is already in salt. As expected, only other file with changes is postgresql.conf. Diff is:

[postgresql.conf.txt](https://github.com/open-contracting/deploy/files/4763945/postgresql.conf.txt)

It's a messy diff, which mainly seems to be down to the pg-10 to pg-11 upgrade. Next step is to work out what changes are actually important and compare that to the in server check you suggest. 

For ref, here is the original postgresql.conf as a text file so github is happy: 
[postgresql.conf.original.txt](https://github.com/open-contracting/deploy/files/4763972/postgresql.conf.original.txt)


ps. I've read before uploading and I think there is no security issues in making public but if you disagree delete this comment

odscjames:
[newpostgres.txt](https://github.com/open-contracting/deploy/files/4764062/newpostgres.txt)

Here's what the diff gives. It's untested. I haven't had time to check against in server yet. 

But so far I'd question if we still need WAL, and I'd see if I can update 10->11 inthe 2 old settings without causing issues, just to avoid confusion later.

odscjames:
I have now checked the settings in the database. Everything listed there is either set by the client, the diff file I posted above or the original distribution config so that is fine.

I guess for procedure would be to put the diff file in  /etc/postgresql/11/main/conf.d/kingfisher.conf and have salt update that. When we deploy we could put the postgresql.conf file back to the original distribution by hand. (I did a basic test in VM)

[...]

PR #162, opened 2020-06-12
kingfisher: Add Postgres Specific settings

odscjames:
https://github.com/open-contracting/deploy/issues/144

### DIFF
--- a/salt/postgres/ocdskingfisher_postgres.conf
+++ b/salt/postgres/ocdskingfisher_postgres.conf
@@ -1,0 +1,43 @@
+# THIS FILE IS MANAGED BY SALT - DO NOT EDIT MANUALLY
+
+
+listen_addresses = '*'
+
+shared_buffers = 16GB
+
+temp_buffers = 128MB
+
+# Caution: it is not advisable to set max_prepared_transactions nonzero unless
+# you actively intend to use prepared transactions.
+max_prepared_transactions = 100
+
+work_mem = 10MB
+
+maintenance_work_mem = 8GB
+
+max_stack_depth = 6MB
+
+bgwriter_delay = 100ms
+
+effective_io_concurrency = 1000
+
+max_parallel_workers_per_gather = 4
+
+wal_level = logical
+
+wal_buffers = 16MB
+
+wal_writer_delay = 1000ms
+
+max_wal_senders = 10
+
+max_replication_slots = 10
+
+log_timezone = 'localtime'
+
+cluster_name = '10/main'
+
+stats_temp_directory = '/var/run/postgresql/10-main.pg_stat_tmp'
+
+timezone = 'localtime'
+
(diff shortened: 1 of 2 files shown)


odscjames:
[Kingfisher: Changes to postgresql.conf are not in Salt](https://trello.com/c/ssC6B4tA/222-kingfisher-changes-to-postgresqlconf-are-not-in-salt)

odscjames:
Prepared pull request as requested in https://github.com/open-contracting/deploy/issues/144

The plan is when we deploy this, we set postgresql.conf back to the original by hand - see issue for a copy.

I note the following problems with the current Salt / this P.R.:

- salt/postgres11.sls file has a file name that suggest it is for any postgres server but it already has kingfisher specific stuff in it. So I continued by adding stuff here.
- salt/postgres11.sls file prob has sequence errors on first install - It might add /etc/postgresql/11/main/pg_hba.conf first then install the package and the package might alter it? 
- salt/postgres11.sls does not restart server if it's config files change ...
- .... but that is complex, because if the postgres server might restart we need to make sure all the tasks in salt/ocdskingfisherprocess.sls that do database things will happen after the restart (or before, just not during. After probably makes sense.) Also, you need to give the postgres server some time to restart after issuing the command before trying to do DB actions or it might not be ready yet.  

In the interests of clearing up a bad situation on the server, unblocking other work that I understand relies on this and as the current Salt has these issues anyway, I'll put this forwards for consideration now and suggest we turn those things into issues afterwards - but open to discussion, of course.

odscjames:
Also note the end result of this should be that no setting actually changes on the DB server - it just makes what is there managed by salt. In the issue I've commented on some things we might want to change but again lets do those later.

jpmckinney:
Thanks, @odscjames. @dogsbody @RobHooper: I'll be away next week, but I'm happy for you to take this forward in your work on setting up a replica.

RobHooper:
@odscjames I'm happy the postgres configuration changes - in the sense that this commits what is currently configured on the server. (I haven't double checked the values themselves as I plan to do further performance tuning later down the line.)

I agree with the problems you identified, as part of the postgres-replica build I've been refactoring the postgres11.sls file so a fix is inbound. 

One fix I've pulled forward here is to restart postgres after config changes. I think this is needed to address a couple of risks:
- Reducing the chance of the "active" postgres configuration and the one in salt becoming out of sync 
- A broken config being deployed and not realised until something else restarted the service

The second change is renaming the config file, this is what my refactored code will be using so it will make the implementation of that easier.

Bjwebb:
I'm taking over Open Data Services' Kingfisher dev work from @odscjames for a few weeks.

Those changes look good. I think we're ready to merge and deploy this pull request. @RobHooper, do you want to do that, or shall I?

RobHooper:
I agree, ready to merge and deploy. Happy for either of us to do this.

Bjwebb:
Would be great if you could do it.

RobHooper:
:+1: I'll merge and deploy this tomorrow morning (UK Time)

Bjwebb:
Looks like this didn't get deployed? Are you able to pick this up next week?

RobHooper:
Sorry for the delay Ben, I tried to deploy twice and was blocked by people using Kingfisher. Deployed now.

Bjwebb:
Thanks, makes sense.

PR #163, opened 2020-06-19
Core install improvements

RobHooper:
Ready for the postgres-replica build I have been reviewing and expanding the base install (core.sls) this pull request is just for changes to the core install.

I have tested these changes on both a fresh Ubuntu 18.04 server and an Ubuntu 18.04 server with the current master core.sls install.

To summarise the changes:
New features:
- Install and configure postfix for email
- Configuring swappiness (swap at 90% memory usage by default, reduced to 60% on low memory systems)
- swap size is now calculated based on memory
- Installing NTP
- Setting UTC timezone
- Disabling ubuntu release notifications
- Disabling Ubuntu MOTD
- Disabling X11forwarding in SSHD
- Timestamp in the history

Changes:
- core is now a directory adding modularity
- email contacts are now variables rather than being hardcoded
- user customisations are in a seperate file (installing htop, vim etc)
- swap uses fallocate which is much faster than dd
- fail2ban is now in its own directory consolidated with configuration files
- Locale set up is in its own file

Removed:
- unattended upgrades configuration (this will be done with the maintenance setup)

The swap change will only be applied if `/swapfile` is missing so it won't be applied to the current servers. It will need a manual kick to be updated with the following commands:
```
swapoff /swapfile                         
rm /swapfile                         
```


jpmckinney:
Looking good!

salt/top.sls
```diff
@@ -3,6 +3,16 @@
...
+    - core.sshd
+    - core.swap
+    - core.mail
+    - core.fail2ban
+    - core.locale
+    - core.ntp
+    - core.motd
+    - core.apt
+    - core.systemd
+    - core.customisation
```
jpmckinney: Does the order matter for these? If not, alphabetical order is preferred.

salt/core/locale.sls
```diff
@@ -0,0 +1,20 @@
+## Locale
+add_system_locale_en_gb_utf8:
+  # This line may be in the file but commented out.
+  # However we use file.append here instead of file.replace because we can't guarantee each Linux Distro build will comment out the lines in the same way.
+  # Could use replace with "- append_if_not_found: True"
```
jpmckinney, line 4: Should we just do that?

salt/core/mail.sls
```diff
@@ -0,0 +1,51 @@
...
+admin:
+  alias.present:
+    - target: root
+postmaster:
+  alias.present:
+    - target: root
+
+# pillar[[admin_username]] :
+#  alias.present:
+#    - target: root
```
jpmckinney, line 43: Why is this commented out? Can it be deleted?

salt/core/mail.sls
```diff
@@ -0,0 +1,51 @@
+# Install and configure email software
+postfix:
+  pkg:
+    - installed
+  service:
+    - running
```
jpmckinney, line 6: Elsewhere we use `service.running`. Should we use it here for consistency?

Other services also `enable: True`. Do we not want postfix to be enabled at boot time?

salt/core/sshd.sls
```diff
@@ -0,0 +1,29 @@
...
+    - name: /etc/ssh/sshd_config
+    - pattern: "^#?PasswordAuthentication .*"
+    - repl: PasswordAuthentication no
+
+# Technically not needed but setting for posterity
+force root ssh keys:
+  file.replace:
+    - name: /etc/ssh/sshd_config
+    - pattern: "^#?PermitRootLogin.*"
+    - repl: "PermitRootLogin without-password"
```
jpmckinney: If not needed, can we delete it?

salt/core/ntp.sls
```diff
@@ -0,0 +1,23 @@
...
+  service.dead:
+    - enable: False
+
+ntp:
+  pkg:
+    - installed
+  service.running:
+    - enable: True
+    - watch: # restart if we change the config
+      - file: /etc/ntp.conf
```
jpmckinney: For fail2ban and ssh, we `reload: True`. Does ntp support reload? Or do we definitely want to restart?

salt/core/fail2ban/init.sls
```diff
@@ -0,0 +1,24 @@
...
+    - sls: core.mail
+  watch:
+    - file: /etc/fail2ban/jail.local
+
+/etc/fail2ban/jail.local:
+  file.managed:
+    - source: salt://core/fail2ban/jail.local
+
+# In the future I'd like to whitelist both admin and server IP's as well.
+# grains.get ip_interfaces
```
jpmckinney, line 23: Let's create an issue and delete the comment. Comments aren't visible for task prioritization.

salt/core/apt.sls
```diff
@@ -0,0 +1,8 @@
+# Disable ubuntu release notifications
+/etc/update-manager/release-upgrades:
+  file.replace:
+    - name:
```
jpmckinney: Is this line needed?

salt/top.sls
RobHooper: I've made these alphabetical. Any order that is needed it is defined using the "require" arguments in the state files themselves.

salt/core/locale.sls
RobHooper: I've updated this file to use file.replace with append_if_not_found. 
This is slightly better as it removes the risk of it getting set twice.

salt/core/mail.sls
RobHooper: I have deleted this

salt/core/mail.sls
RobHooper: I've updated this to use `service.running` and added `enable: True`.

salt/core/ntp.sls
RobHooper: NTP doesn't support reloads. It restarts really quickly so restarting isn't a problem.

salt/core/apt.sls
RobHooper: `- name:` line removed.

salt/core/fail2ban/init.sls
RobHooper: I've removed this comment. I will come to you with a number of future enhancements I've collected together at a later point. Then we can decide if we want to make this into an issue.

salt/core/sshd.sls
RobHooper: We'd like to keep this setting so that it is clear to users that you cannot log in to root without a password.
As I said in my comment we're already technically doing this by also setting with `PasswordAuthentication no`. 

There are two reasons for doing this, firstly it adds an extra layer to the security if `PasswordAuthentication` is toggled back on, the second reason is that while `PermitRootLogin` is set to either `no` or `without-password` we can monitor it with our intrusion detection software.

I think we should update the comment to show this, how do you think it would be best to phrase it?

jpmckinney: I think your 2nd-3rd sentences explain it well in a comment :) Feel free to copy-edit as desired, update and merge.

RobHooper: Comment improved.

PR #169, opened 2020-07-28
workflows: Functional tests of live cove-ocds site, that are run with cron

Bjwebb:
Moved from
https://github.com/OpenDataServices/opendataservices-deploy/blob/413d1495b2bce4bf520608cb1fa506382028e699/.travis.yml#L20


Bjwebb:
[Fix daily run of cove-ocds functional tests against the live site](https://trello.com/c/wieroUOx/68-fix-daily-run-of-cove-ocds-functional-tests-against-the-live-site)

Bjwebb:
Still todo:
- [ ] Update the notification email(s)
- [ ] (After merge) Set up Travis cron for master (needs to be done through the web UI)

@jpmckinney What email(s) do you want to be notified of this? (The Travis default is whoever made the last commit, which is not so useful for this). I'm happy to leave our [existing email](https://github.com/open-contracting/deploy/pull/169/files#diff-354f30a63fb0907d4ad57269548329e3R16) in, as that's still useful for us, and multiple emails can be specified.

jpmckinney:
Hmm, we use GitHub Workflows now, not Travis anymore (there are open issues to migrate the remaining cove-ocds and cove-oc4ids repos). Can you change to GitHub workflows? I think you can use `on.schedule`: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

Bjwebb:
Yes, I'll do that. The travis file was just a copy/paste anyhow, so I've not really wasted any time in creating it.

Bjwebb:
I've moved it to GitHub Actions, this needs merging onto master before the cron will run.

Looks like notifications for cron workflows go to whoever made the last commit, and this isn't configureable. One alternative I've seen discussed is set up a custom email step, using your own email service or smtp server. Another option would be to move this to its own github repo, and ensure the most recent commit is authored by whatever emails we want to receive notficiations. @jpmckinney Do you have any thoughts on this?

jpmckinney:
I prefer the email step, e.g. using this action: https://github.com/marketplace/actions/send-email

We can have it trigger only if a step failed.

.github/workflows/cove-ocds-live-tests.yml
```diff
@@ -0,0 +1,34 @@
+name: CI
+on: 
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 0 * * *'
+env:
+  BASEDIR: https://raw.githubusercontent.com/open-contracting/standard-maintenance-scripts/master
```
jpmckinney, line 7: We don't need these lines – I assume they were copy-pasted.
```suggestion
```

.github/workflows/cove-ocds-live-tests.yml
```diff
@@ -0,0 +1,34 @@
+name: CI
```
jpmckinney, line 1: We use "CI" when doing continuous integration on the repo itself.

```suggestion
name: OCDS Data Review Tool
```

Bjwebb:
@jpmckinney Makes sense. Do we have an appropriate account for sending the email via, or is that something we need to set up?

jpmckinney:
Yes, you can search for noreply@noreply.open-contracting.org or otherwise find the credentials in `deploy/salt/private/prometheus-server-alertmanager/conf-alertmanager.yml`

Bjwebb:
I've added an email step with those details. I've tested it with "on: push", but the final commit will now only run on the cron (which only runs once this is in master).

Looks like I can't turn off the regular nofitications to the author of the last commit.

Bjwebb:
@RobHooper Just a heads up that if you made the most recent commit, you might get a notification email from this ("Run failed: OCDS Data Review Tool live tests"). This is safe to ignore as we'll get notified too.

PR #175, opened 2020-08-11
Adding support for new maintenance set up repo

RobHooper:
This PR adds in a new private repo containing a number of salt scripts that set up maintenance. 
You can see these salt scripts here: https://github.com/open-contracting/dogsbody-maintenance

@JimACarter, please can you review the linked salt code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Private repositories
 /pillar/private
 /salt/private
+/salt/maintenance
 # Files created by script/setup
 /salt-config/master.d
 /salt-config/pki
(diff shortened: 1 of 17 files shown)


JimACarter:
Notes from reviewing this and then discussing with Rob:
- Fix code blocks formatting on github in README
- Can we support case insensitive patching variables - to support Manual?
- Expand comment around salt core lm-sensors in README to make it clearer that the user needs to pull the modules out of lm-sensors output and add them to the pillar data
- Set a consistent value for variables that are unset, I'd probably opt for None where possible
-- Then might not need comparison in the if statements
- For software raid, check that mdadm is running and alerting us/root?

RobHooper:
Thanks @JimACarter , I've fixed the issues you've flagged up. 
Regarding software raid monitoring thanks for flagging this up. 
I've removed my software raid monitoring script as it isn't needed, the mdadm tool does this and is set up as part of the pre-salt server build. I've changed the README to reflect this.
I don't see us installing mdadm ourselves to I've haven't set up salt to install / configure it. If we do configure it we'll add it then.

@jpmckinney, just checking you're happy with this pull request.

One remaining job is to update pillar data for each server to match their maintenance set up but I'm not expecting any more changes in salt.

jpmckinney:
Looks good to me! I haven't reviewed the other repo closely, but happy to move forward.

PR #178, opened 2020-08-26
Update ijson

aguilerapy:
To resolve https://github.com/open-contracting/deploy/commit/54c3738009d58f295dcaf05662d3a0f7df26ade6 and https://github.com/open-contracting/deploy/commit/eb45e6a7305911d9cd1bfd35f6e8c5fa2160f2a4.

### DIFF
--- a/salt/ocdskingfishercollect/scrapyd-requirements.txt
+++ b/salt/ocdskingfishercollect/scrapyd-requirements.txt
@@ -1,4 +1,4 @@
-ijson>=3
+ijson>=3.1.1
 jsonpointer
 jsonref
 jsonschema


jpmckinney:
@yolile Out of curiosity, why is ijson>=3.1.1 required? Did we encounter a bug in earlier versions?

yolile:
to use the "use_float" argument (introduced in 3.1.1 version) in  https://github.com/open-contracting/kingfisher-collect/blob/5ab8eebba13a736a9216f593c624fc830c7adb4e/kingfisher_scrapy/spiders/portugal_base.py#L24
otherwise, the float values are converted to Decimal objects.

Issue #176, opened 2020-08-18
Postgres enhancements to reduce replication issues and speed up restoration


RobHooper:
These are improvements I noted while investigating and restoring broken replication between the kingfisher servers this morning.

* We are currently storing 10 wal logs, this is fine most of the time, however last night there was a huge amount of data written to kingfisher-process1 and it failed to keep up.
* Setting up postgres "replica slots" that should force the master server to hold onto wal logs until the replica gives it the OK.
* We are archiving old wal logs - which I was able to use to restore kingfisher-replica1 from. This directory is very big and we should start clearing it down to speed up restores. - I propose keeping the last week worth of wal logs.
* We could also start proactively copying archived wal logs between servers, then postgres will be able to automatically restore from them should this happen again.

RobHooper:
Increasing the `max_wal_size` will also help with item 1. This already has an issue #158

jpmckinney:
Sounds good to me. Would just replica slots (point 2) be enough?

RobHooper:
Replica slots may well be enough to keep replication working, however I'm not sure yet how/if we can automate setting this up.
The others are more nice to haves, changing the stored wal logs + wal log size is a quick win as they are just config settings.
Clear down the wal archive is the next priority as it could lead to the disk slowly filling up.

RobHooper:
Raised PR #182 which addresses point 3 / archiving old wal logs

Regarding replica slots. I have looked into moving them into salt and decided to leave this as a manual step for now.
Most of the changes for replica slots are custom to the two servers we've set it up on and I'm not sure of the best way to make this a "template" and reusable for future servers at the moment. I suggest picking this up again when new postgres replicas come online.

I've documented this set up at my end. 
For reference, these are the postgres replica slot specific commands I ran:
```
# On kingfisher-replica1
echo "primary_slot_name = 'replica1'" >> /var/lib/postgresql/11/main/recovery.conf

# On kingfisher-process1
psql
select * from pg_create_physical_replication_slot('replica1');
```
The rest of recovery.conf is autogenerated by pg_basebackup - [see docs](https://ocdsdeploy.readthedocs.io/en/latest/deploy/postgres.html)


Replica slots have been working well so I see no need for any additional fixes (proactively copying wal logs to the replica).

jpmckinney:
@RobHooper Sounds good. Can you document the location of `recovery.conf` to allow for manual inspection?

> I've documented this set up at my end.

[...]

PR #182, opened 2020-09-09
Adding script to clear down Postgres WAL archive

RobHooper:
Adding script to clear down WAL archive on Postgres replica master servers after 7 days.

Ref issue #176

### DIFF
--- a/docs/deploy/postgres.rst
+++ b/docs/deploy/postgres.rst
@@ -111,29 +111,27 @@ You will configure a master server and a replica server.
 
          pg_lsclusters
 
-#. Optional steps
+#. It is recommended to enable replication slots:
 
-   #. Enable automatic WAL archive restoration on the replica
-
-      .. code-block:: bash
-
-         echo "restore_command = 'cp /var/lib/postgresql/11/main/archive/%f %p'" >> /var/lib/postgresql/11/main/recovery.conf
-
-   #. Enable replica slots on the replica server
+   #. On the replica server:
 
       .. code-block:: bash
 
          echo "primary_slot_name = 'example_unique_identifier'" >> /var/lib/postgresql/11/main/recovery.conf
          service postgresql restart
 
-   #. Enable replica slots on the master server
+   #. On the master server:
 
       .. code-block:: bash
 
-         sudo su - postgres
-         psql
-         > select * from pg_create_physical_replication_slot('example_unique_identifier');
+         su - postgres
+         psql -c "SELECT * FROM pg_create_physical_replication_slot('example_unique_identifier');"
 
+#. (Optional) Enable automatic WAL archive restoration on the replica server:
+
+   .. code-block:: bash
+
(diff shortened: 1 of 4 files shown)


jpmckinney:
Should `pg_archivecleanup` be used instead? https://www.percona.com/blog/2019/07/10/wal-retention-and-clean-up-pg_archivecleanup/

Here it is used from the replica: https://stackoverflow.com/a/16951547/244258 as mentioned here: https://www.postgresql.org/docs/12/warm-standby.html#STANDBY-SERVER-SETUP Here it is used by a shell script on the master, which is discouraged by some: https://stackoverflow.com/a/22965933/244258

On this page https://wiki.postgresql.org/wiki/Streaming_Replication and elsewhere, it's suggested to simply turn off archiving for streaming replication:

```
# Enable WAL archiving on the primary to an archive directory accessible from
# the standby. If wal_keep_segments is a high enough number to retain the WAL
# segments required for the standby server, this is not necessary.
```

So, perhaps the best option is to turn off archiving, and make sure `wal_keep_segments` is high enough?

RobHooper:
I'd like to keep WAL archiving because it has already helped us restore the replica twice without requiring a full database dump / replica reset. 
We _shouldn't_ need it now that replica slots makes sure enough WAL segments are kept in `pg_wal`/`pg_xlog`. But I like having archiving as a fallback in case anything goes wrong.

As you've found `pg_archivecleanup` is a particularly good tool to clear down WAL files on replica servers, but the WAL archives are not proactively copied over so there isn't any need for it on there (copying archives to replica is something I considered and decided to not implement as mentioned on #176 ). 

We can use `pg_archivecleanup` to keep master in check but as these links mention, you still need a wrapper script to find and pass the correct arguments. We'd be swapping the script in my PR to run something like...
```find /var/lib/postgresql/11/main/archive/ -mtime +7 -printf '%f\n' | sort -r | head -1 | xargs pg_archivecleanup /var/lib/postgresql/11/main/archive/ ```
IMO this has the same results as the current proposed script so isn't worth doing. 


I think the discouragement you saw regarding this was around deleting files in `pg_wal`/`pg_xlog` and this is indeed bad as Postgres manages this directory. We are not doing this, instead we are looking to delete files from the `archive` directory which is a custom directory we've made and is only referred to in this [one config line](https://github.com/open-contracting/deploy/blob/868643da90a3f514ec7c12e135add821c4e93a06/salt/postgres/configs/kingfisher-process1-postgres.conf#L49).

jpmckinney:
I'm fine with a 'belts-and-suspenders' approach. Can you add a comment in Salt explaining that:

> "Replica slots should make sure that enough WAL segments are kept in pg_wal/pg_xlog. 7 days of archives serves as a fallback in case anything goes wrong.

RobHooper:
I've added these comments around maintaining the archive directory into the salt script.
I've also updated postgres.rst pointing to `recovery.conf` and adding some further docs around set up.

jpmckinney (approved):

nutza249949:
thanks

PR #191, opened 2020-10-06
Install and configure server firewalls

RobHooper:
This PR uploads a new firewall management script to `/home/sysadmin-tools/bin/firewall.sh`
This script uses iptables to manage the firewall.

By default the firewall script locks down everything except SSH. 
Other service ports are then opened up when the software is installed.

This script sets up port knocking however I want to double check this before implementing on the servers.

Port knocking means that we can restrict access to SSH, only opening it up when someone visits port 8255 before hand.
This is good for users with dynamic IP addresses. 
This is security through obscurity, in that anyone can work out how to port knock by looking at the code in this repo, however most script kiddies etc won't, so it'll keep the chaff out. 

My concerns with port knocking are:
1. I believe there are a number of users who SSH into Kingfisher to query Postgres, they would all need informing of this new system.
I can write some nice docs to help with this.
Alternatively we could leave this feature disabled on the kingfisher server. 

2. By default port knocking gives you a 60 second time frame to connect, this could interrupt slow running salt deployments.
We could increase the timeout to address this. 

Users with static IPs would be whitelist through and continue working at normal.


RobHooper:
I have tested this deployment on a Postgres and Apache environment.

jpmckinney:
@RobHooper Great! I haven't read the PR carefully, but trust it works. The only thing I'm unsure about is port-knocking.

For PostgreSQL, when it uses `sslmode=require`, does that require port-knocking or not?

Otherwise, I guess we will need to port-knock before every salt-ssh command?

I'm happy to merge after there is documentation on port-knocking.

jpmckinney:
For salt-ssh, we might use the pre-flight config to portknock: https://docs.saltstack.com/en/latest/topics/ssh/roster.html

RobHooper:
1. Port knocking isn't currently required for Postgres. It currently opens port 5432 to the world when salt installs Postgres.
I've added functionality to have Postgres locked down to specific IPs in the firewall script but it isn't currently used. 

2. With the current PR we would indeed need to port knock for every salt-ssh command.
I've added a simple bash script that does this.
``` ./script/portknock ex1.open-contracting.org ```

I've tested `ssh_pre_flight` which does exactly what we're looking for but there are a few issues with it. 
* Salt expects this to be set to a single script and doesn't let you use arguments (like my above example).
* This needs to be the full path to the script rather than a relative path, which since we've probably all got the repo deployed in different places, doesn't work for us. 

3. As I put in my first message this is "security through obscurity" which isn't really security. If we disable port knocking there is one less step between attackers and your server but that step is  public on this git repo. 

4. If you want to continue with port knocking, which wiki page should I put the below docs against?
Perhaps a new userguide page for SSH usage, as well as adding linking from the salt-deployment pages ([like this one](https://ocdsdeploy.readthedocs.io/en/latest/develop/get_started.html) )

```
SSH access to our servers is locked down using port knocking.
port pnocking means that we can keep SSH locked down by default, blocking script kiddies and other bad internet traffic from scanning port 22 but still enable port 22 on a per IP basis when it is required.

You can port knock by sending any traffic to port 8255, the easiest is done by browsing to it http://ex1.open-contracting.org:8255 , the page will time out but it still works. (replacing ex1.open-contracting.org with the server hostname you're connecting to)

This whitelists your IP for 60 seconds letting you connect.
Once you're connected, you'll become an established connection and stay whitelisted until you log out.

If you are using Salt there is a script to do this for you
 ./script/portknock ex1.open-contracting.org 

If you have a static IP address we can whitelist access from it so you don't need to port knock. 
```

**Comment by jpmckinney** [comment]
Hmm, but if we do something like `salt-ssh '*'` then we'll need to port-knock each server, which is a pain – unless I'm missing some easy alternative?

I wrote and committed a script, so we'd run commands with `./run.py` instead of `salt-ssh`.

On macOS, I have to use `-G` not `-w` for the `nc` command. Does that also work on Linux?

If the script works, then I can make a commit to replace all relevant occurrences of salt-ssh with `./run.py`.

**Comment by jpmckinney** [comment]
In case a misconfiguration causes the port to be permanently closed, Bytemark has a web console, and I think Hetzner's [Rescue System](https://docs.hetzner.com/robot/dedicated-server/troubleshooting/hetzner-rescue-system/) can be used, correct? We can add docs about that, maybe to `troubleshoot.rst`

Actions to do in this PR or in a follow-up PR:

* [ ] Add `nc` port-knock to deploy-docs.sh
* [ ] Add short how-to on use/index.rst to knock before ssh
* [ ] Update all occurrences of `ssh`
* [ ] Update all occurrences of `salt-ssh`

**Comment by RobHooper** [comment]
@jpmckinney, I've tweaked run.py to use the python socket library for port knocking, please can you test it works on your system?
`nc -G` doesn't work on our system at all so I think this is a good alternative. 

I've also updated the shebang `env` command to run `python3` because `python` still points at 2.7 for me.


Regarding misconfiguration concerns, you are correct there are fallback rescue systems with Bytemark and Hetzner. I'll add some docs around how we expect these to work. 
We also have the DBT office IP hardcoded in the firewall script as a fallback.


FYI, my current local salt install has a bug causing new macros to not work. I've had to downgrade to the previous salt major version (3001).

**Comment by jpmckinney** [comment]
So Salt stopped working with 3002? I've updated requirements.txt on master to use 3001.

run.py works!

Do you want to do the checklist items as part of this PR, or in new issues? Let me know if I should take care of any of them.

**Comment by RobHooper** [comment]
I think our repo code is compatible with salt 3002, I think it is just a bug in the latest client update (at least on our Ubuntu systems).
Hopefully it's be fixed upstream soon - https://github.com/saltstack/salt/issues/58896.

I've updated `deploy-docs.sh` and documented the port knocking set up. 
Either of us can do the `salt-ssh` to `run.py` update, it makes sense to add it to this PR.

File: docs/use/ssh.rst [hunk 1]
```diff
@@ -0,0 +1,43 @@
+Connecting to servers with SSH
+==============================
+
+.. note::
+    Before you can SSH into the OCP servers you first need to upload your SSH public key.
+    `See notes on salt install page. <https://ocdsdeploy.readthedocs.io/en/latest/develop/get_started.html?highlight=ssh#add-public-key-to-remote-servers>`_.
```
jpmckinney, line 6: Add a `.. _add-public-key:` anchor on `get_started.rst` and then use:

```rst
See :ref:`add-public-key`.
```

jpmckinney (commented):
We might want to link from the "Connect to servers" section of `kingfisher.rst` to the new `ssh.rst` page.

Please go ahead with the remaining updates of occurrences of `ssh` and `salt-ssh`.

docs/use/ssh.rst
RobHooper: Fixed in commit aacfdedd99b3400c3a80aa60250ef01c18831650

RobHooper:
@jpmckinney docs changes committed. I think we're all good to proceed with the deployment. I'll do this one morning next week.

Have we informed users who SSH in that port knocking is about to be added? I'm not sure which mailing list is best to ping.

jpmckinney (approved):

jpmckinney:
Please ping servers-status and kingfisher-status. Thanks!

RobHooper:
> So Salt stopped working with 3002? I've updated requirements.txt on master to use 3001.

@jpmckinney the salt developers have fixed this bug upstream now so we can roll back this change and use any version of salt :)

jpmckinney:
Great! I've removed that constraint.

PR #194, opened 2020-10-13
Changing uwsgi configuration salt and adding docs

RobHooper:
This PR changes the salt based uwsgi configuration management.

One of the main things we found is that the uwsgi configs are quite messy because they are being changed depending on the environment in a number of different ways.
My aim in this PR is to remove the complexity making it easier to develop in the future.

So, most of these changes are for clarity, improving variable names and structure.

This PR also adds better default settings for a number of uwsgi variables.

Alongside the code changes I've added some docs around setting up new uwsgi and Django applications
I haven't put in wiki mark up yet, I wanted to check the content before doing so.

@JimACarter , please can you double check the new uwsgi setting values in pillar.
@jpmckinney , please can you review the code.


jpmckinney:
I was also looking into alternatives to uwsgi, since many places report bad performance / unnecessary complexity.

Do you have any experience with mod_wsgi? I read about it here: https://www.digitalocean.com/community/tutorials/django-server-comparison-the-development-server-mod_wsgi-uwsgi-and-gunicorn

JimACarter:
I'm happy with the pillar settings.

Regarding the switching to mod_wsgi / gunicorn.  We haven't heavily tested and compared different python webservers.  The one we use to run our warboard and a few other internal tools is uWSGI, the data manipulation they're doing isn't causing nearly as much memory usage though.
AFAIK uWSGI is as lightweight as we're going to get but I think we're agreed that the code is the primary factor in the memory usage we're seeing. We need a webserver that we're able to configure to mange this. I think we're able to do this with uWSGI.
If everyone is agreed that a different webserver would be easier to configure (I find that uWSGI does a good job at listing the args but a worse one at explaining some) I think that's a valid reason to switch.

jpmckinney:
Let's stick with uwsgi since it's what's familiar, and since we're unlikely to reach the performance limits where mod_wsgi, gunicorn, etc. might prove better.

Happy to merge once docs are tidied.

RobHooper:
@jpmckinney I've just updated the docs. Please can you review.

You mentioned uWSGI adds "unnecessary complexity", I realise that this PR adds a a lot more Jinja markup into the uWSGI configs which does make things much messier - however this allows us to have greater control from pillar. 
This is changeable if it is a problem.

jpmckinney:
> You mentioned uWSGI adds "unnecessary complexity", I realise that this PR adds a a lot more Jinja markup into the uWSGI configs which does make things much messier - however this allows us to have greater control from pillar.
> This is changeable if it is a problem.

I was reporting others' opinions on uwsgi – though even just comparing its [documentation](https://uwsgi-docs.readthedocs.io/en/latest/) to others, it's a bit of a mess. But it's working well enough, and it's what we know, so there's no strong reason to switch to mod_wsgi or gunicorn at this time.

I think this PR actually makes things simpler, since we now have an INI file that is better controlled by Pillar!

jpmckinney (approved):

PR #199, opened 2020-11-02
Updating elasticsearch to version 7.x and updating configs to match

RobHooper:
ElasticSearch version 6.x goes end of life on the 20 Nov 2020. 

This pull request updates ES to version 7.x and updates the configs to match. 
I've also made a couple of improvements to how the apt key is stored and consolidating the custom configs together.

Below is the update process. These commands need to be ran in this order on standard-search (unless otherwise specified):

```
rm /etc/apt/sources.list.d/elasticsearch.list
# On your local system, salt-ssh state.apply
systemctl stop elasticsearch.service
apt-get update && apt-get dist-upgrade
# Yes accept defaults (two of them are overwritten by salt anyway)
# On your local system, salt-ssh state.apply
```

Finally we'll want to kick off a re-index of the docs, please can you advise here?
I can see the [deploy-docs.sh script](https://github.com/open-contracting/deploy/blob/master/deploy-docs.sh), is there any docs on how this is called?

We'll also need to update the [standard-search](https://github.com/open-contracting/standard-search) repo as the readme says ES version 6.

### DIFF
--- a/salt/standard-search.sls
+++ b/salt/standard-search.sls
@@ -2,12 +2,11 @@ include:
   - django
 
 elasticsearch:
-  cmd.run:
-    - name: wget -nv -O - https://packages.elasticsearch.org/GPG-KEY-elasticsearch | apt-key add -
   pkgrepo.managed:
     - humanname: Elasticsearch
-    - name: deb https://artifacts.elastic.co/packages/6.x/apt stable main
+    - name: deb https://artifacts.elastic.co/packages/7.x/apt stable main
     - file: /etc/apt/sources.list.d/elasticsearch.list
+    - key_url: https://packages.elasticsearch.org/GPG-KEY-elasticsearch
   pkg.installed:
     - pkgs:
       - apt-transport-https
@@ -26,7 +25,7 @@ elasticsearch:
 
 /etc/default/elasticsearch:
   file.managed:
-    - source: salt://etc-default/elasticsearch-standard-search
+    - source: salt://standard-search/elasticsearch-defaults
     - template: jinja
 
 /etc/elasticsearch/jvm.options:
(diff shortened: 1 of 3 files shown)


jpmckinney:
> Finally we'll want to kick off a re-index of the docs, please can you advise here?
> I can see the deploy-docs.sh script, is there any docs on how this is called?

The deploy-docs.sh script sends a request to standard-search, which basically web-scrapes the documentation and replaces the index.

Do we need to do that? My understanding is that Elasticsearch offers several options for reindexing: https://www.elastic.co/guide/en/elasticsearch/reference/current/reindex-upgrade.html

jpmckinney (approved):
Looks good!

I take it the config files are copies of e.g. https://github.com/elastic/elasticsearch/blob/7.10/distribution/src/config/jvm.options with the variables filled in.

jpmckinney:
I've removed the version number from the standard-search repo, where it's not really relevant (the ES API methods we use are very stable and haven't changed between at least 6 and 7).

RobHooper:
Ah, looking at the docs you've linked a reindex might not be needed at all. 
> Elasticsearch can read indices created in the previous major version.

The elasticsearch configs are indeed copies of the upstream standard ES 7 configs keeping the few customisations that had been made.

jpmckinney:
Sounds good. By the time there is ES8, we will have created/updated the indices through our regular operations.

Would you like to deploy? Let me know once done, and I'll test that everything still works as expected (basically: doing a search from https://standard.open-contracting.org/latest/en/).

RobHooper:
I'll deploy now quick. :+1:

jpmckinney:
Search works! I think I need to update something in standard-search or the JS that calls it to fix:

> Search finished, found **[object Object]** page(s) matching the search query.

jpmckinney:
Found the breaking change: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#hits-total-now-object-search-response

Fixed now!

PR #220, opened 2020-12-16
Adding docs around replication failover and recovery

RobHooper:
@jpmckinney Just a little PR adding more docs for Postgres. 

I noted that the `postgres-replica` DNS record currently has an hour TTL.
Ideally, we'd reduce this so that failover happens quicker. This is currently a CNAME so it would make sense to change `postgres-replica` to an A record and leave the `ocp05` record as is.


docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
 .. code-block:: sql
 
    SELECT '16712'::regclass;
+
+
```
jpmckinney, line 171: Omit extra newlines.

```suggestion
```

jpmckinney:
> I noted that the `postgres-replica` DNS record currently has an hour TTL.

Can you start another PR to edit the recommended TTLs in https://ocdsdeploy.readthedocs.io/en/latest/deploy/create_server.html#create-dns-records? The original TTLs for A and AAAA records were 1 day 3328d1a8e3a2692379299cce46a0c5c3a2b2048b. I'm fine with whatever TTL as long as there's an explanation.

> Ideally, we'd reduce this so that failover happens quicker. This is currently a CNAME so it would make sense to change `postgres-replica` to an A record and leave the `ocp05` record as is.

Why should postgres-readonly.kingfisher.open-contracting.org be an A record instead of a CNAME?

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+Fixing replication:
+
+# Copy over source replica log archives
+
+  .. code-block:: bash
+
+     service postgres stop
+     sudo su - postgres
+     rsync -avz postgres@process1.kingfisher.open-contracting.org:/var/lib/postgresql/11/main/archive/ /var/lib/postgresql/11/main/archive/
```
jpmckinney, line 211: The postgres user on the replica server has SSH access to the postgres user on the process server, and without having to port knock?

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+     exit
+     service postgres start
+
+# Monitor the replica logs, you'll see it recovering from the WAL files.
+
+  .. code-block:: bash
+
+     tail -f /var/log/postgresql/postgresql-11-main.log
+
```
jpmckinney, line 220:
```suggestion
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+If everything goes wrong and you can fall back to rebuilding the replica from source, see :ref:`postgres setup docs<pg-setup-replication>`.
```
jpmckinney, line 222: What's an example of "everything goes wrong"?

```suggestion
If all else fails, then you can fallback to rebuilding the replica. See :ref:`pg-setup-replication`.
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+  .. code-block:: bash
+
+     service postgres stop
+     sudo su - postgres
+     rsync -avz postgres@process1.kingfisher.open-contracting.org:/var/lib/postgresql/11/main/archive/ /var/lib/postgresql/11/main/archive/
+     exit
+     service postgres start
+
+# Monitor the replica logs, you'll see it recovering from the WAL files.
```
jpmckinney:
```suggestion
#. Monitor the replica logs. You should see messages about recovery from WAL files.
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+   Wait for the DNS updates to complete. This will depend on the records TTL value.
+
+Done.
+
+
+Fixing replication:
+
+# Copy over source replica log archives
+
+  .. code-block:: bash
```
jpmckinney, line 207:
```suggestion
  .. code-block:: bash
```

This should be 3-space indentation.

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
 .. code-block:: sql
 
    SELECT '16712'::regclass;
+
+
+
+Recovering the replica
+----------------------
```
jpmckinney, line 174: Use predicates, not gerunds.
```suggestion
Recover the replica
-------------------
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+In the case that replication breaks or the replica server goes offline these are the steps to recover.
+There are two parts to this, mitigating the downtime and fixing replication.
```
jpmckinney, line 177: Put paragraphs on a single line.

```suggestion
If replication breaks or the replica server goes offline, you must recover the replica, in two stages: mitigate the downtime, and fix the replication.
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+Mitigating downtime:
```
jpmckinney, line 179:
```suggestion
Mitigate downtime
~~~~~~~~~~~~~~~~~
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+# Enable public access to postgres on kingfisher-process1
```
jpmckinney:
```suggestion
# Enable public access to the PostgreSQL service on the main server
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+   Modify ``pillar/kingfisher.sls`` setting the following pillar value
```
jpmckinney, line 183: Put this after the code block. We start with generic instructions, then give specific examples.

```suggestion
   For example, for the ``kingfisher-process`` target, modify the ``pillar/kingfisher.sls`` file.
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+   Log into GoDaddy
+
+   Point the postgres-readonly DNS endpoint at kingfisher-process1
+
+   Wait for the DNS updates to complete. This will depend on the records TTL value.
+
+Done.
+
+
+Fixing replication:
```
jpmckinney, line 203:
```suggestion
Fix replication
---------------
```

RobHooper:
> Why should postgres-readonly.kingfisher.open-contracting.org be an A record instead of a CNAME?

Sorry I think this statement of mine was wrong in hindsight. We can change the TTL of the postgres-replica CNAME without changing the ocp05 records.


I'll create a new Git issue to discuss TTLs further.


I think this PR is good to merge.

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+   Modify ``pillar/kingfisher.sls`` setting the following pillar value
+
+   .. code-block:: yaml
+
+      postgres:
+        public_access: True
+
+# Deploy changes to kingfisher-process1
+
```
jpmckinney: We can omit newlines between consecutive list items. Link to other docs.

```suggestion
#. :doc:`Deploy the main server<../../deploy/deploy>`
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+# Update DNS
+
+   Log into GoDaddy
+
+   Point the postgres-readonly DNS endpoint at kingfisher-process1
+
+   Wait for the DNS updates to complete. This will depend on the records TTL value.
+
+Done.
```
jpmckinney: Use numbered lists for all individual instruction steps, not paragraphs. No need to say "done". The end of a list means "done".

```suggestion
#. Update DNS records:

   #. Login to `GoDaddy <https://sso.godaddy.com>`__
   #. If access was delegated, open `Delegate Access <https://account.godaddy.com/access>`__ and click the *Access Now* button
   #. Open `DNS Management <https://dcc.godaddy.com/manage/OPEN-CONTRACTING.ORG/dns>`__ for open-contracting.org
   #. Update the replica's CNAME record to point to the main server's A record: for example, point ``postgres-readonly`` to ``kingfisher-process1``
   #. Wait for the changes to propagate, which depends on the original TTL value
```

docs/maintain/databases.rst
```diff
@@ -167,3 +167,56 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+
+Fixing replication:
+
+# Copy over source replica log archives
```
jpmckinney:
```suggestion
#. Copy WAL archives from the main server to the replica server. For example:
```

docs/maintain/databases.rst
RobHooper: Fixed

docs/maintain/databases.rst
RobHooper: Fixed

docs/maintain/databases.rst
RobHooper: Fixed

docs/maintain/databases.rst
RobHooper: Fixed

docs/maintain/databases.rst
```diff
@@ -167,3 +167,55 @@ To get the table related to a ``pg_toast_*`` table, take the number after ``pg_t
...
+
+#. Copy WAL archives from the main server to the replica server. 
+
+   Replace ``example.open-contracting.org`` below with the postgres source server hostname.
+
+   .. code-block:: bash
+
+      service postgres stop
+      sudo su - postgres
+      # Copy the SSH key for the postgres user to the postgres user on example.open-contracting.org
```
jpmckinney, line 209: The postgres user has no SSH key, though.

jpmckinney:
I added a note to explain the SSH requirements.

Do we want to set up those SSH keys in Salt?

RobHooper:
Yes, that's a good idea - I've added a block in `postgres/main.sls` to manage postgres authorized_keys.

I've been trying to refer to the kingfisher-process1 postgres instance as the "source".
It might be nice to rename the `main.sls` file to `source.sls` or `replica-source.sls` to keep everything the same.

jpmckinney:
Hmm, I think "main" is used more frequently than "source" in our code/docs. Okay to stick with "main"?

pillar/kingfisher.sls
```diff
@@ -34,6 +34,10 @@ ssh:
...
     - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC8VRLRwHk2qjF/68XQcxabz6whVbkZQKG4mP5KUUnMPWdC1/xF8cgbWaQc34s/9yaoeV03bzbQJU7DCvOZwIWnUnLbPvlAuzsHVbzYPooYbvIGg2CK9LK6AuWli7bCgTJ31FxNbt0yRmF1dEsWLr2M2XiL/rRuF25kA+XyWsRKSmcuggUlY5H+kC1v8TcNOYJFw5SKy7pq6Wa5MrErbGzyWsm2Rq0NhYMQecvTdcN5d7/wsZUXOMwmMddsf1C3gtHCawIR7h8787g+AywC2ggEvR47Mv4Ed1je8Vc1En1tFpznmE93gZ154am/eDFkK3oRWQFPqyBZg0azk+570WiJnHPCe9qJ74c4bygH+B3bQ0XQKiEJOmEyK1lIW8ze/0DrknbBIRY14836eepbu7Xa/JVK+HOZk51tT4Zo2Zb0xRo5BwE3RfRwMAHNEA3/L3bDD4lpn3y379ryW5MstVU2yy4mZEep+RQOj0kRFjiYV5zzmOQbBUKeQt4FFjgXZQdVTQAqbXGB9CAnpJ69L78fgQwD0rSmJUsK+2kCPClAhtCQx50mXelhUJnIwsP/LyqtQAu73M9PhPpnrHRz1DNCobVO31rqLM0kag9GjNYAtizn1Jviozs03k0FK40vMU1pGSkiuCYl3hyv+GtU5wbSUjOoH8uD3MaJphZJzOUS/w==
+  postgres:
+    # postgres@replica1.kingfisher.open-contracting.org
+    - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDDNn6x0P4ZBrZeTTugOqGWo9EFZWuFVxi0oRQkpbd4g92Yu6EgKj3xJlFcgn14oi58/3w6nHSytj/V4vCXNLPkOla3/7XGpbop5hllsdJftAZJgbuHjgcKjIXVlw8QVrvt39PgQVqyXLIyHK0lpn8QUCUMfRH2axOfWscvtmhZpKCR3DYKjGknCuRhxgjhCblVEEbzoXJaxFhQxfTj4GmGzEcK/3H/SaNDXnvDlRrcjQZ2PjJ41cNGfV/hxX9GIANRm151PqJ3ftESv78c8N1KmK3dA9SRDDImFXGIy/6KPRtzQ+CWgEZkYey0O7vcuytyy8z7FionbGyLW4qBmlD9
+
```
jpmckinney, line 40:
```suggestion
```

jpmckinney (commented):
Was the private key manually created? We can either store it in deploy-salt-private, or we can add a step to the docs for setting up the replication server.

I suggested removing an extra newline.

RobHooper:
It was manually created, I don't mind either approach.
I prefer adding `ssh-keygen` in the docs to prevent key re-use. On the other hand it is nice having everything in salt.

jpmckinney:
> It was manually created, I don't mind either approach.
> I prefer adding `ssh-keygen` in the docs to prevent key re-use. On the other hand it is nice having everything in salt.

We can do both: Add `ssh-keygen` in the docs for new replicas, and store keys in Salt once replicas created.

RobHooper:
@jpmckinney Should be addressed in the latest commits.

jpmckinney (approved):
Nice! I didn't know about `contents_pillar`. I'll use that for another key instead of salt/private/keys.

Issue #193, opened 2020-10-09
smartmon.sh: Should be owned by root


jpmckinney:
@JimACarter had written:

> I also wanted to mention that having the /home/prometheus-client/node-exporter-textfile-collector-scripts/smartmon.sh file owned by prometheus-client and run as root via cron gives easy privilege escalation for the prometheus-client user.

From shared document:

> It should be owned by root and in directories that are owned by root as it's run by root via cron.  Potentially worth discussing moving node-exporter-textfile-collector-scripts to /opt or a preferred location.

PR #224, opened 2020-12-18
Move smartmon.sh to a directory owned by root

JimACarter:
Closes https://github.com/open-contracting/deploy/issues/193


jpmckinney (approved):
Looks good!

PR #227, opened 2020-12-23
git repo update script enhancement

RobHooper:
Small PR, this change makes sure that the update script is loaded into memory before executing.

Bash scripts by default execute each line as they go, this meant that if an update changed the update script it would run the git pull and update itself, then start executing the new script half way through.

This will be needed when we remove repos like "salt/private".

### DIFF
--- a/script/update
+++ b/script/update
@@ -3,23 +3,27 @@
 
 set -xeuo pipefail
 
-git checkout master
-git pull --rebase origin master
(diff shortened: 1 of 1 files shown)


jpmckinney (approved):

jpmckinney:
Fascinating, I didn't know that!

PR #248, opened 2021-02-17
Allow access to postgres from appbeat IPs

### LINKED ISSUE #166: postgres status checking during deployments
You can easily deploy a bad postgres configuration because salt checks `systemctl status postgres` which shows the postgres service as running even if `pg_lsclusters` shows it is actually down. 

We hope that any issues will be picked up in testing, but it would be better if salt checked and alerted users that something is wrong when the deployment happens.

### LINKED ISSUE #205: PostgreSQL: Allow connections from Appbeat
https://www.appbeat.io/probes/ipv4
https://www.appbeat.io/probes/ipv6

RobHooper:
Enables AppBeat monitoring, closes #205 
Also fixes Postgres not identifying if the service is broken after a service restart, closes #166

### DIFF
--- a/salt/postgres/files/pg_hba.conf
+++ b/salt/postgres/files/pg_hba.conf
@@ -30,3 +30,14 @@ hostssl replication     replica         {{ ip }}/32             md5
 hostssl replication     replica         {{ ip }}/128            md5
   {%- endfor %}
 {%- endif %}
+
+{% if salt['pillar.get']('maintenance:enabled') == True %}
+# Appbeat monitoring
+  {%- for ip in salt.cmd.run('curl -sS https://www.appbeat.io/probes/ipv4').split() %}
+hostssl nodatabase      nouser         {{ ip }}/32             md5
+  {%- endfor %}
+
+  {%- for ip in salt.cmd.run('curl -sS https://www.appbeat.io/probes/ipv6').split() %}
+hostssl nodatabase      nouser         {{ ip }}/128             md5
+  {%- endfor %}
+{%- endif %}
--- a/salt/postgres/files/pg_hba.conf
+++ b/salt/postgres/files/pg_hba.conf
@@ -21,23 +21,23 @@ hostssl	all		all		::/0			md5
 {%- if 'replica_ipv4' in pillar.postgres %}
 # Allow SSL connections from replicas (IPv4).
   {%- for ip in pillar.postgres.replica_ipv4 %}
-hostssl	replication	replica		{{ "{:18}".format(ip+"/32") }}	md5
+hostssl	replication	replica		{{ "{:18}".format(ip + "/32") }}	md5
   {%- endfor %}
 {%- endif %}
 {%- if 'replica_ipv6' in pillar.postgres %}
 # Allow SSL connections from replicas (IPv6).
   {%- for ip in pillar.postgres.replica_ipv6 %}
-hostssl	replication	replica		{{ "{:42}".format(ip+"/128") }}	md5
+hostssl	replication	replica		{{ "{:42}".format(ip + "/128") }}	md5
   {%- endfor %}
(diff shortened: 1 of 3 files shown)


salt/postgres/files/pg_hba.conf
```diff
@@ -1,32 +1,43 @@
...
-hostssl all             all             ::/0                    md5
+hostssl	all		all		0.0.0.0/0		md5
+hostssl	all		all		::/0			md5
 {%- endif %}
 
 {%- if 'replica_ipv4' in pillar.postgres %}
 # Allow SSL connections from replicas (IPv4).
   {%- for ip in pillar.postgres.replica_ipv4 %}
-hostssl replication     replica         {{ ip }}/32             md5
+hostssl	replication	replica		{{ "{:18}".format(ip+"/32") }}	md5
```
jpmckinney, line 24: Please add spaces around the `+` operators.

salt/postgres/files/pg_hba.conf
```diff
@@ -1,32 +1,43 @@
...
   {%- endfor %}
 {%- endif %}
 {%- if 'replica_ipv6' in pillar.postgres %}
 # Allow SSL connections from replicas (IPv6).
   {%- for ip in pillar.postgres.replica_ipv6 %}
-hostssl replication     replica         {{ ip }}/128            md5
+hostssl	replication	replica		{{ "{:42}".format(ip+"/128") }}	md5
```
jpmckinney, line 30: Why is the width 18 above but 42 here?

salt/postgres/files/pg_hba.conf
```diff
@@ -1,32 +1,43 @@
...
+  {%- endfor %}
+{%- endif %}
+
+{% if salt['pillar.get']('maintenance:enabled') == True %}
```
jpmckinney, line 34:
```suggestion
{%- if salt['pillar.get']('maintenance:enabled') %}
```

salt/postgres/init.sls
```diff
@@ -16,6 +16,9 @@
   {% if pillar.postgres.get('replica_ipv6') %}
     {{ set_firewall("REPLICA_IPV6", pillar.postgres.replica_ipv6|join(' ')) }}
   {% endif %}
+  {% if salt['pillar.get']('maintenance.enabled', True) %}
```
jpmckinney, line 19: The current code would default to True, but I think we want to test if True.

```suggestion
  {% if salt['pillar.get']('maintenance:enabled') %}
```

jpmckinney (commented):
Some comments.

So the issue with #166 is just that the service needs to be more precisely identified?

jpmckinney:
I've updated the issue description to auto-close the related issues.

salt/postgres/files/pg_hba.conf
RobHooper: 18 is the maximum size of an IPv4 address with the subnet - IPv6 is longer at 42.

This is just to make the `pg_hba` file easier to read once it's generated keeping the columns aligned.

RobHooper:
> So the issue with #166 is just that the service needs to be more precisely identified?

Yes indeed - it turns out that the `postgresql` service is a "dummy" service that just calls the specific version service in the background. This update calls that specific service directly fixing our issues.
This is why the `postgresql` service shows it running regardless of if it is or not, because it doesn't monitor the postgres daemon at all.

salt/postgres/files/pg_hba.conf
RobHooper: Done

salt/postgres/files/pg_hba.conf
jpmckinney: Aha, I hadn't noticed it was for different IP versions :)

jpmckinney (approved):

PR #254, opened 2021-03-19
PostgreSQL backups

RobHooper:
Not ready to merge in yet, just raising PR to track remaining tasks and for some discussion.

This pull request sets up pgbackrest for Postgres backups. 
Backups are configured on both the kingfisher-process1 and kingfisher-replica1 servers. Backups can then be initiated on the source server, where pgbackrest then SSH's into the replica to backup the data itself, only using the source server for a few required files. 
This also improves our processes around setting up replication.

To do:
- [x] AWS Credentials permissions issues
- [ ] Duplicate replica user pillar info
- [x] Set up SSH access so that process1 can connect into replica1
- [x] Update Documentation (for the replication set up and restoration processes)


RobHooper:
@jpmckinney please can you provide a second opinion on a few salt problems. :)

1. I need to get this pillar data `pillar.get postgres:users:replica` on the kingfisher-replica server, my only thought is to duplicate it and set it in `private.kingfisher_replica`. 
Do you have any suggestions around this?

2. I need to whitelist the kingfisher-process server IPs on kingfisher-replica. 
In an ideal world I would add the new IP to `pillar.get firewall:ssh_ipv4` and it would merge with the current IPs (`pillar/private/common.sls`) however when I do this salt overwrites the variable rather than merging, so that doesn't seem to be an option.
I could define the admin IPs again for kingfisher-replica but again duplicate data isn't ideal.
One other thought is that we create a new variable `firewall:additional_ssh_ipv4` and merge in the firewall script. 
Thoughts?

Also FYI, I will be coming to you regarding the AWS credentials issues, I'm just trying to narrow the problem down first (happy that the salt side works OK - I have had this working in testing into my own AWS account).

jpmckinney:
> 1. I need to get this pillar data `pillar.get postgres:users:replica` on the kingfisher-replica server, my only thought is to duplicate it and set it in `private.kingfisher_replica`.
>    Do you have any suggestions around this?

We can move it to a new Pillar file (maybe `kingfisher_common.sls`), and include it for each server in the top file?

> 2. I need to whitelist the kingfisher-process server IPs on kingfisher-replica.
>    In an ideal world I would add the new IP to `pillar.get firewall:ssh_ipv4` and it would merge with the current IPs (`pillar/private/common.sls`) however when I do this salt overwrites the variable rather than merging, so that doesn't seem to be an option.
>    I could define the admin IPs again for kingfisher-replica but again duplicate data isn't ideal.
>    One other thought is that we create a new variable `firewall:additional_ssh_ipv4` and merge in the firewall script.
>    Thoughts?

Yes, I did something similar with `pillar.ssh.admin` being added to an optional `pillar.ssh.root` in `salt/lib.sls`. We can either name the new variable `additional_ssh_ipv4` or rename the original variable `ssh_ipv4_common` (it's in `common.sls`), which might better express its semantics as being global.

> Also FYI, I will be coming to you regarding the AWS credentials issues, I'm just trying to narrow the problem down first (happy that the salt side works OK - I have had this working in testing into my own AWS account).

OK, let me know. I also created a Backblaze account (free for now as no data is stored 🙂 ).

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
 
 The configuration file will be in the ``/etc/postgresql/11/main/conf.d/`` directory on the server (for PostgreSQL version 11).
 
+.. _pg-setup-backups:
+
+Set up backups
+--------------
+
+We use pgbackrest to create and manage backups stored in the cloud.
```
jpmckinney, line 58:
```suggestion
We use `pgBackRest <https://pgbackrest.org>`__ to create and manage offsite backups.
```

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
...
+Salt will install and configure pgbackrest if ``postgres:backup`` is defined in pillar.
```
jpmckinney, line 59: Use correct capitalization of pgBackRest throughout, unless referring to the command.

```suggestion
Salt will install and configure pgBackRest if ``postgres:backup`` is defined in Pillar data.
```

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
...
+        backup:
+          # The configuration file for pgbackrest, this is loaded from ``salt/postgres/files/pgbackrest/``.
+          config_name: kingfisher-process1
+          # Unique ID to store backups under. pgbackrest also refers to this as the "stanza".
+          identifier: kingfisher
+          # Concurrent processes for run pgbackrest with (backup speed vs CPU usage).
+          # Optional.
+          process_max: 4
+          # Backup bucket region.
+          region: eu-west-1
```
jpmckinney, line 75: I think we tend to use us-east.

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
...
+          # API Access Key.
+          s3_key: redacted
+          # API Secret Key.
+          s3_key_secret: redacted
+          # Total full backups to store.
+          total_full_backups: 4
+
+   .. note::
+
+      These options support any s3 compatible storage systems including AWS, BackBlaze and more.
```
jpmckinney, line 89:
```suggestion
      These options support any S3-compatible storage, including AWS and BackBlaze.
```

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
...
+
+   .. note::
+
+      Incremental backups are taken daily (storing only the changes since the last full backup).
+      Full backups are taken weekly, currently this runs on Sunday.
+      So if ``total_full_backups`` is set to 4, backups will be stored for four weeks.
+
+#. Create Stanza.
```
jpmckinney, line 97:
```suggestion
#. Create stanza.
```

docs/develop/update/postgres.rst
```diff
@@ -48,14 +50,118 @@ Add your configuration
...
+
+   If this backup stanza has already been created you can skip this step.
+
+   .. code-block:: bash
+
+      su - postgres
+      pgbackrest stanza-create --stanza=example
+
+
```
jpmckinney, line 106: Extra newline (occurs elsewhere).

```suggestion
```

docs/develop/update/postgres.rst
```diff
@@ -103,96 +219,23 @@ To configure a main server and a replica server:
...
+         mkdir /var/lib/postgresql/11/main
+         pgbackrest --stanza=example --type=standby restore
 
    #. Switch to the ``root`` user and start the PostgreSQL service.
 
       .. code-block:: bash
 
          exit
-         service postgresql start
+         systemctl status postgresql
```
jpmckinney, line 235: This starts the PostgreSQL service?

docs/develop/update/postgres.rst
```diff
@@ -29,6 +29,8 @@ To enable public access, update the server's Pillar file:
   postgres:
     public_access: True
 
+.. _postgres-add-configuration:
```
jpmckinney, line 32: We use `pg-` instead of `postgres-` for all except `postgres-public-access`, so we can change this and `postgres-public-access` to use `pg-`.

docs/maintain/databases.rst
```diff
@@ -339,6 +339,60 @@ Show running queries:
 
 See the `pg_stat_activity <https://www.postgresql.org/docs/11/monitoring-stats.html#PG-STAT-ACTIVITY-VIEW>`__ table's documentation.
 
+.. _pg-recover-backup:
+
+Restore from backup
+-------------------
+
+Postgres is being backed up onto offsite cloud storage, backups and restoration are managed by `pgbackrest <https://pgbackrest.org/>`__.
```
jpmckinney, line 347:
```suggestion
PostgreSQL databases are backed up offsite. Backup and restoration are managed by `pgBackRest <https://pgbackrest.org/>`__.
```

docs/maintain/databases.rst
```diff
@@ -339,6 +339,60 @@ Show running queries:
...
+
+You can either remove this data if it is irrelevant:
+
+.. code-block:: bash
+
+   su - postgres
+   rm -rf /var/lib/postgresql/11/main
+   mkdir /var/lib/postgresql/11/main
+
+Or if it is related data for instance when you are rolling back following a change or fixing corrupted data.
```
jpmckinney, line 386: I don't understand this sentence fragment.

salt/postgres/backup.sls
```diff
@@ -0,0 +1,16 @@
+include:
+  - postgres
+
+# Require official PostgreSQL repo as they provide a newer version of pgbackrest.
+pgbackrest:
+  pkg.installed: []
```
jpmckinney, line 6: Why an empty array?

docs/develop/update/postgres.rst
RobHooper: Happy to use one of the us-east regions instead of eu-west.
These three regions cost the same ($0.023 per GB).

It is a bit more expensive ($0.0245) but eu-central-1 (Frankfurt) is the closest to the Hetzner servers (hosted in Falkenstein and Helsinki). So for quicker backups and restorations this region would be better.

salt/postgres/backup.sls
RobHooper: Salt handles the empty array the same as:
```
pgbackrest:
  pkg.installed
```
Because of the additional `require` under this you need to set `pkg.installed` to something otherwise it would be invalid yaml.

If it isn't clear, we could use this instead
```
pgbackrest:
  pkg.installed: pgbackrest
  require: ... 
```

jpmckinney: Aha. See `mail.sls` for the preferred style.

docs/develop/update/postgres.rst
jpmckinney: I guess we should decide whether we will be using Backblaze instead?

RobHooper:
@jpmckinney Shifting the discussion on using Backblaze here. 
I have tested backing up and restoring a ~40MB postgres database into AWS S3 (eu-west-1 region) and to Backblaze b2 (eu-central region). The server was hosted in Bytemark. I had pgbackrest set up with two max connections.

These are the times I recorded:
* Backup to S3 - 26.099s
* Backup to Backblaze - 6m28.759s
* Restore from S3 - 17.979s
* Restore from Backblaze - 59.217s

The costs for both are:
* AWS - $0.023 per GB ( or $0.0245 for eu-central-1)
* Backblaze - $0.005 per GB

As you can see Backblaze is significantly slower at both backups and restorations (especially backups).
We're backing up __a lot more__ data on Kingfisher than I used in this test so this is a big concern.

Based on the above, I think we should continue with AWS and see how long a full backup takes before we set up Backblaze.

jpmckinney:
@RobHooper Sounds good. Thanks for testing!

RobHooper:
@jpmckinney Please can you update the AWS IAM user to have the following permissions policy:

``` json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::ocp-backup"
            ],
            "Condition": {
                "StringEquals": {
                    "s3:prefix": [
                        "",
                        "kingfisher"
                    ],
                    "s3:delimiter": [
                        "/"
                    ]
                }
            }
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::ocp-backup"
            ],
            "Condition": {
                "StringLike": {
                    "s3:prefix": [
                        "kingfisher/*"
                    ]
                }
            }
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:PutObject",
                "s3:GetObject",
                "s3:DeleteObject"
            ],
            "Resource": [
                "arn:aws:s3:::ocp-backup/kingfisher/*"
            ]
        }
    ]
}
```

jpmckinney:
@RobHooper The current policy seems to be less restrictive. What is the intended outcome of the change?

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "s3:PutObject",
                "s3:GetObject",
                "s3:ListBucket",
                "s3:DeleteObject"
            ],
            "Resource": [
                "arn:aws:s3:::ocp-backup",
                "arn:aws:s3:::ocp-backup/*"
            ]
        }
    ]
}
```

RobHooper:
The problem is that the current IAM policy is incompatible with pgbackrest.
I agree that it looks less restrictive so I'm not sure why it doesn't like it. 
I *think* IAM might be handling the "ListBucket" permissions differently between the two.

I've tested running a backup with the policy I shared above and it works well - this is also the policy that is shared on the official documentation.

jpmckinney:
Can you link to the example?

RobHooper:
@jpmckinney https://pgbackrest.org/user-guide.html#s3-support

docs/develop/update/postgres.rst
RobHooper: This was a typo. Fixed now.

docs/maintain/databases.rst
```diff
@@ -339,6 +339,50 @@ Show running queries:
...
+.. code-block:: bash
+
+   pgbackrest restore --stanza=example --delta
+
+Restore specific backup by timestamp:
+
+.. code-block:: bash
+
+   pgbackrest restore --stanza=example --set=20210315-145357F_20210315-145459I --delta
+
```
RobHooper, line 374: @jpmckinney I've rewritten this paragraph in the docs. Aiming to address _this comment https://github.com/open-contracting/deploy/pull/254#discussion_r605267107_.

How does the below read?

jpmckinney:
I've updated the IAM policy.

docs/develop/update/postgres.rst
jpmckinney: Happy to use an AWS EU region.

docs/maintain/databases.rst
jpmckinney: Yes, it's clearer!

docs/develop/update/postgres.rst
```diff
@@ -103,14 +105,179 @@ Add your configuration
...
+
+#. Create an S3 bucket and API Keys.
+
+   .. note::
+
+      pgBackRest supports any S3-compatible storage, including AWS and BackBlaze.
+
+   If you are using AWS you will need to `create an S3 Bucket<https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html>`__ and `set up an IAM user<https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html>`__.
+
+   This is the IAM permissions policy we are using for Kingfisher backups.
```
jpmckinney, line 124: Link to https://pgbackrest.org/user-guide.html#s3-support

salt/postgres/backup.sls
```diff
@@ -0,0 +1,32 @@
...
+
+/etc/pgbackrest/pgbackrest.conf:
+  file.managed:
+   - makedirs: True
+   - user: postgres
+   - group: postgres
+   - source: salt://postgres/files/pgbackrest/{{ pillar.postgres.backup.config_name }}.conf
+   - template: jinja
+
+{%- if salt['pillar.get']('postgres:backup:enabled','') == True %}
```
jpmckinney, line 19: The following is sufficient:

```suggestion
{%- if salt['pillar.get']('postgres:backup:enabled') %}
```

salt/postgres/init.sls
```diff
@@ -66,6 +66,10 @@ postgresql-reload:
     - mode: 640
     - watch_in:
       - service: postgresql
+{% if salt['pillar.get']('postgres:backup:identifier') %}
```
jpmckinney, line 69:
```suggestion
{% if salt['pillar.get']('postgres:backup:enabled') %}
```

salt/top.sls
```diff
@@ -71,6 +71,9 @@ base:
   'I@postgres:configuration':
     - postgres
 
+  'I@postgres:backup':
```
jpmckinney, line 74:
```suggestion
  'I@postgres:backup:enabled:true':
```

salt/postgres/backup.sls
RobHooper: I have made this change, I also renamed this variable to show it's actual purpose (now called `cron_enabled`).

RobHooper:
The last few commits should improve the separation between the three different set ups we need: Postgres installed without pgBackRest, pgBackRest installed and not running, pgBackRest installed and running automatically. 

This second category is particularly important. Kingfisher will start in category two, we need pgBackRest installed and configured but we want to find out how long backups take and how large they are before we enable the cron. 

We also will have kingfisher-replica which needs pgBackRest installed to restore replication but will never have backups running.

pillar/kingfisher_backups.sls
```diff
@@ -0,0 +1,10 @@
+postgres:
+  backup:
+    enabled: False
+    identifier: kingfisher
+    total_full_backups: 4
+    process_max: 4
+    s3_bucket: ocp-backup
+    s3_endpoint: s3.eu-west-1.amazonaws.com
+    region: eu-west-1
```
jpmckinney, line 8: Rename to `s3_region`.

pillar/kingfisher.sls
```diff
@@ -109,3 +106,5 @@ postgres:
     - 148.251.183.230
   replica_ipv6:
     - 2a01:4f8:211:de::2
+  backup:
+    config_name: kingfisher-process1
```
jpmckinney, line 110: To follow the same pattern as for postgres. Need to make corresponding changes.

```suggestion
    configuration: kingfisher-process1
```

pillar/kingfisher_backups.sls
```diff
@@ -0,0 +1,10 @@
+postgres:
+  backup:
+    enabled: False
+    identifier: kingfisher
```
jpmckinney, line 3: Should `identifier` be `stanza`?

salt/postgres/init.sls
```diff
@@ -66,6 +66,10 @@ postgresql-reload:
...
+    - require:
+      - pkg: pgbackrest
```
jpmckinney, line 71: Should this be `sls: postgres.backup` ?

jpmckinney:
> The last few commits should improve the separation between the three different set ups we need: Postgres installed without pgBackRest, pgBackRest installed and not running, pgBackRest installed and running automatically.

Thanks! Yes, I had misunderstood the logic of `enabled`. `cron_enabled` is clearer.

pillar/kingfisher_backups.sls
RobHooper: The `identifier` is indeed the `stanza` - I felt that `identifier` made the purpose clearer to admins.
Happy with either :)

pillar/kingfisher_backups.sls
```diff
@@ -0,0 +1,10 @@
...
+    total_full_backups: 4
```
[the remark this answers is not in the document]
RobHooper: Happy swapping `total_full_backups` to `retention_full` - I've made this change in a new commit.

I think just `path` would be confusing, in pgBackRest there are multiple `*-path` variables (`pg1-path`, `repo1-path`). `repo_path` keeps it clear that this is the path on the destination.

pillar/kingfisher_backups.sls
jpmckinney: I find it easier when a single word is used for a given concept. `stanza` is unusual, but it's what pgbackrest uses.

pillar/kingfisher_backups.sls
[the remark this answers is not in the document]
jpmckinney: Aha, find with `repo_path` then.

docs/develop/update/postgres.rst
RobHooper: Updated docs to `s3_region: eu-central-1` matching kingfisher settings.

pillar/kingfisher_backups.sls
RobHooper: Updated variable name to `stanza`

salt/postgres/init.sls
RobHooper: I agree, `sls: postgres.backup` is a bit better because it gets the pgBackRest configuration in place as well.
I've made this change in a new commit.

RobHooper:
@jpmckinney I think we're good to merge into main and deploy. Please can you confirm?

My last commit fixes the last few issues, including Postgres "replica" user duplication in pillar.

Also noting, I have some unpushed commits to `pillar/private` that need committing at the same time as this is deployed.

jpmckinney (approved):

jpmckinney:
Yes, looks great!

PR #263, opened 2021-04-22
covid19: 262-create-env-for-covid-19-procurement-explorer-admin

bikramtuladhar:
* [x] Created .env jinja template in salt/covid19/files/.env
* [x] Created salt function with file.managed with pillar context for .env file


bikramtuladhar:
related issue #263

jpmckinney:
I added commits and merged from the command line.

PR #265, opened 2021-05-05
Enabling regular postgres backups

RobHooper:
I manually ran a pgBackRest Postgres backup this morning it took roughly 2 hours 15 minutes to complete and compressed the backup size down to only 500GB. 
I am very happy with both of these stats :smile: 

Based on this information, I expect AWS S3 to cost roughly $50 - $60 per month.
We can potentially reduce this cost by moving to S3 infrequent access or possibly Glacier.
I don't have good visibility on the number of requests the backup process made which may change this recommendation.

Salt is currently configured to keep the last four full backups (`retention_full`).

This PR enables daily incremental backups with a weekly full backup (running on Wednesday morning).

### DIFF
--- a/pillar/kingfisher.sls
+++ b/pillar/kingfisher.sls
@@ -108,4 +108,10 @@ postgres:
     - 2a01:4f8:211:de::2
   backup:
     configuration: kingfisher-process1
-    # cron_enabled: True # Disabled to test backup timing
+    process_max: 8
+    cron: |
+        MAILTO=root
+        # Daily incremental backup
+        15 05 * * 0-2,4-6 postgres pgbackrest backup --stanza=kingfisher
+        # Weekly full backup
+        15 05 * * 3 postgres pgbackrest backup --stanza=kingfisher --type=full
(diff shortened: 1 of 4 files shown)


RobHooper:
@jpmckinney Just double checking you're happy with these costings before enabling? :)

jpmckinney:
What’s the cost of Infrequent Access, and the pros/cons? My understanding is it could be about half the cost, as long as we aren’t frequently accessing the backups.

jpmckinney:
Sounds good! Let's start with IA then.

jpmckinney (approved):

RobHooper:
@jpmckinney Sorry about this, I thought it was possible to use other storage classes but I can't find the setting. 
I think I have mixed up pgbackrest and another backup tool.

The only mention of S3 IA support I can find is in this Git issue where they are dubious if it will save money.
https://github.com/pgbackrest/pgbackrest/issues/371

Because we are only storing objects for 4-5 weeks, we need to set the S3 objects storage type when we upload it so until pgbackrest supports this it looks like this isn't an option.

Anyway, I will proceed with the PR as is, storing 4 full backups in S3 standard costing roughly $60 per month.

jpmckinney:
That's fine. The question mark was around the number of requests, and that thread makes it seem like there can actually be a fair number of requests, which would end up bringing IA to the same cost as Standard.

Issue #233, opened 2021-01-06
Consider setting limits to /var/log/journal


jpmckinney:
For example, It's taking up 2.4GB on the docs server (10%).

We can set `SystemMaxUse=` https://www.freedesktop.org/software/systemd/man/journald.conf.html

After doing that, we can run `journalctl --vacuum-size=` once.

RobHooper:
We have seen multiple disk space alerts on the live.docs server since January. 
I have manually set `SystemMaxUse` on this server to help prevent the alerts, I will create a PR to formalise this soon.

PR #267, opened 2021-05-18
Configuring journald in salt

RobHooper:
Setting journald log storage limits as we've seen disk space issues on the live.docs server.


jpmckinney (approved):
#233 can then be closed?

RobHooper:
Yep #233 can be closed once this is deployed.

PR #280, opened 2021-07-19
Updating SSH Key Comments

RobHooper:
PR adds comments to all SSH Keys 

We find SSH Key comments very useful for debugging.
For instance in the past, we have seen alert for the covid19 server, we can see that someone was logged in and from the logs we can see the hash of their SSH key.
We can cross-check this hash with the keys in authorized_keys (ssh-keygen -lf /home/covid19admin/.ssh/authorized_keys)
Currently we need to double check the server keys against the ones in GitHub to find the user, if we had SSH comments set up it would make this step much easier.

I have tried to standardise comments for users to:
> FirstName LastName (CompanyShortHand, UserRole, AuthoritativeLink) email

Or for servers:
> username@hostname

There was one instance I wasn't sure what to put as the comment, in `pillar/covid19.sls`. temp-salt-server (deploys covid-19-procurement-explorer-admin).


jpmckinney (approved):
temp-salt-server is a bot used by Young Innovations (YI). Happy for the current comment to be used as the SSH comment, or as you prefer.

PR #281, opened 2021-07-21
Ubuntu 20.04 support

RobHooper:
This PR brings the repo in line with our current Ubuntu 20.04 build best practices.


salt/core/ntp.sls
```diff
@@ -1,4 +1,24 @@
-# Disable pre-installed systemd-timesyncd as we prefer NTP.
+{% if grains['osrelease']|float >= 20.04 %}
```
jpmckinney, line 1: Can't we do a string comparison? If so, make the same change elsewhere.

```suggestion
{% if grains['osrelease'] >= '20.04' %}
```

salt/core/ntp.sls
```diff
@@ -1,4 +1,24 @@
...
+    - repl: 'NTP=0.uk.pool.ntp.org 1.uk.pool.ntp.org 2.uk.pool.ntp.org 3.uk.pool.ntp.org/'
+    - watch_in:
+      - service: systemd-timesyncd
+
+# Catch instances where ntp has been installed previously.
+ntp:
+  service.dead:
+    - enable: False
+
+{% else %}
```
jpmckinney, line 21: Why is the service disabled on earlier Ubuntu versions?

salt/core/ntp.sls
```diff
@@ -1,4 +1,24 @@
...
+systemd-timesyncd:
+  pkg.installed:
+    - name: systemd-timesyncd
+  service.running:
+    - enable: True
+
+/etc/systemd/timesyncd.conf:
+  file.replace:
+    - pattern: "^#?NTP=.*"
+    - repl: 'NTP=0.uk.pool.ntp.org 1.uk.pool.ntp.org 2.uk.pool.ntp.org 3.uk.pool.ntp.org/'
```
jpmckinney, line 12: Why is there a trailing slash?

salt/core/ntp.sls
```diff
@@ -1,4 +1,24 @@
...
+# systemd-timesyncd should be installed by default unless the NTP set up code below has ran against server.
+systemd-timesyncd:
+  pkg.installed:
+    - name: systemd-timesyncd
+  service.running:
+    - enable: True
+
+/etc/systemd/timesyncd.conf:
+  file.replace:
+    - pattern: "^#?NTP=.*"
```
jpmckinney, line 11: This works for the default file (which I found at the bottom of https://dannyda.com/2021/04/27/how-to-change-ntp-server-on-ubuntu-20-04-1-lts-simply-default-contents-of-etc-systemd-timesyncd-conf-on-ubuntu-20-04-1-lts/) but `NTP=` can be stated multiple times, in which case I think this won't have the desired behavior.

salt/core/ntp.sls
```diff
@@ -1,4 +1,24 @@
-# Disable pre-installed systemd-timesyncd as we prefer NTP.
+{% if grains['osrelease']|float >= 20.04 %}
+# systemd-timesyncd should be installed by default unless the NTP set up code below has ran against server.
+systemd-timesyncd:
+  pkg.installed:
+    - name: systemd-timesyncd
+  service.running:
+    - enable: True
```
jpmckinney, line 7: As a good practice, we make the name explicit and require the package.

```suggestion
    - name: systemd-timesyncd
    - enable: True
    - require:
      - pkg: systemd-timesyncd
```

salt/core/sshd/init.sls
```diff
@@ -1,9 +1,19 @@
 # We'll only be using SSH key authentication.
+{% if grains['osrelease']|float >= 20.04 %}
+/etc/ssh/sshd_config.d/customization.conf:
+  file.managed:
+    - source: salt://core/sshd/files/customization.conf
+    - watch_in:
+      - service: ssh_service
+
+{% else %}
```
jpmckinney, line 9: Could we instead add `Include /etc/ssh/sshd_config.d/*.conf` to `/etc/ssh/sshd_config` if not present, and then use the same Salt instructions for both Ubuntu versions?

salt/core/systemd/init.sls
```diff
@@ -1,7 +1,19 @@
 # Stop RemoveIPC killing all processes by a user when they log out.
+{% if grains['osrelease']|float >= 20.04 %}
+/etc/systemd/logind.conf.d/customize.conf:
```
jpmckinney, line 3: We name the others `customization.conf`. Unless it's already taken, we should be consistent.

```suggestion
/etc/systemd/logind.conf.d/customization.conf:
```

salt/core/systemd/init.sls
```diff
@@ -1,7 +1,19 @@
...
+{% else %}
 /etc/systemd/logind.conf:
   file.replace:
     - pattern: "#?RemoveIPC=yes"
     - repl: "RemoveIPC=no"
     - append_if_not_found: True
+{% endif %}
+
+systemd-logind:
+  service.running
```
jpmckinney, line 19: Add the name.

```suggestion
  service.running:
    - name: systemd-logind
```

salt/core/systemd/init.sls
```diff
@@ -1,7 +1,19 @@
 # Stop RemoveIPC killing all processes by a user when they log out.
+{% if grains['osrelease']|float >= 20.04 %}
+/etc/systemd/logind.conf.d/customize.conf:
+  file.managed:
+    - source: salt://core/systemd/files/logind.conf
+    - makedirs: True
+    - watch_in:
+      - service: systemd-logind
 
+{% else %}
```
jpmckinney, line 10: Similar to an earlier comment, I believe Ubuntu 18 already checks /etc/systemd/logind.conf.d/*.conf (based on running `man logind.conf`), so it'd be nice to use the same configuration across versions.

salt/core/ntp.sls
RobHooper: That is a typo from the sed command we normally run. Fixed.

salt/core/ntp.sls
RobHooper: systemd-timesyncd has been pre-installed by ubuntu for a few releases now.
For a long time we preferred using NTP but we have internally agreed that full NTP is overkill for most servers. 
systemd-timesyncd is a simpler implementation of the NTP protocol syncing time to one server rather than against multiple NTP endpoints.

The one thing we need to be careful around is ending up with two different systems being installed both changing the system time.
This PR swaps it around the install to prefer systemd-timesyncd and disable NTP - in the current state of this PR this should only apply to amy01 where NTP has been installed by us already.


Considering your other comments, it'd be nice to use the same configuration across versions, so we could remove the if statement completely here.

jpmckinney: I'm happy with removing the if statement :)

salt/core/systemd/init.sls
RobHooper: Done - using same new configuration across versions.

salt/core/sshd/init.sls
RobHooper: Unfortunately `Include` isn't supported on Ubuntu 18 instances in `sshd_config` (sshd) - it can be used in `ssh_config` but that is for a different purpose.

salt/core/ntp.sls
RobHooper: Yes we can - I didn't think this would work! 

I have removed this specific `if` statement anyway but I have made this string comparison change in other places (`sshd/init.sls`)

salt/core/ntp.sls
RobHooper: I have updated this to put our NTP servers in the `timesyncd.conf.d` directory and can confirm it is taking effect using the following command `timedatectl show-timesync` (this command only works on Ubuntu 20.04).

salt/core/sshd/init.sls
jpmckinney: Oops, I checked the wrong man page. Sounds good then.

jpmckinney (approved):

PR #288, opened 2021-08-16
Limit temp tables to 1TB

RobHooper:
@jpmckinney small PR setting 1TB limit on Postgres temp file.
This only sets the variable on the replica1 server, would we like this set on kingfisher-process1?


jpmckinney (approved):
Just replica is fine, since the main server is more controlled (no individual user accounts).

PR #314, opened 2021-09-28
fix: docker_apps check for target

yolile:
Otherwise a "`Jinja variable 'salt.utils.odict.OrderedDict object' has no attribute 'target'`" error is raised


jpmckinney:
🤔 But all the entries have a `target` and an `env` (in fact, I don't think the if-statemets are needed). This was maybe before the private registry file was properly nested.

yolile:
Maybe. Feel free to test it and remove the if statements if they are not needed anymore.

jpmckinney:
They're now removed.

PR #315, opened 2021-09-29
Registry apache

yolile:


pillar/registry.sls
```diff
@@ -15,8 +15,6 @@ prometheus:
 
 apache:
   public_access: True
-  ipv4: 65.21.93.181
-  ipv6: 2a01:4f9:3b:45ca::2
```
jpmckinney, line 19: Why delete these? See https://ocdsdeploy.readthedocs.io/en/latest/develop/update/apache.html#bind-addresses

yolile:
When I added the new site I got an error about duplicating an existing
configuration. But maybe I was doing something wrong. Feel free to re
introduce them if needed.

Issue #299, opened 2021-09-17
Upgrade PostgreSQL on Redash


jpmckinney:
9.6 will be EOL: https://www.postgresql.org/support/versioning/

We might prefer to switch to this repo https://github.com/getnewdash/setup since the official one isn't updated.

Or, now that we have better support for docker-compose, we might have a more personalized deployment.

Issue #278, opened 2021-07-16
Registry network (netplan) configuration


RobHooper:
The new registry server has been provisioned with two IP addresses 65.21.93.141 and 65.21.93.181, however the server is only configured to use one of them (65.21.93.181).

The DNS for ocp13.open-contracting.org has both of these IPs listed but since one isn't configured this means that there is a good chance deployments will fail. 

The server has also been provisioned with IPv6 however there are no DNS records for this.

@jpmckinney What is this secondary address being used for? do we need it?
If we do, we should reconfigure netplan on this server for both IPs.

We should also create new IPv6 DNS records.

jpmckinney:
Ah, yes, 65.21.93.181 is for data.open-contracting.org and 65.21.93.141 is for flatten.open-contracting.org. Since they each configure their own webserver in Docker, one of which has more complex settings, we figured two IPs would be easier than adding a web server/load balancer in front of both. Please make the necessary configuration changes :)

I don't know the IPv6 addresses (I don't remember seeing them in robot.your-server.de), but I'm happy to add those to DNS.

jpmckinney:
@myroslav I believe you have configured this? cc @sorenabell

myroslav:
@jpmckinney we'd added additional IP on the server.

@anton-shakh-qg can you comment here what actions had you taken to add additional IP?

shakhanton:
1. edited ` /etc/netplan/01-netcfg.yaml`
and added needed IP to the following block

```
...
  ethernets:
    enp9s0:
      addresses:
        - 65.21.93.181/32
        - 65.21.93.141/32
...
```
2. then ran the command `sudo netplan apply`

RobHooper:
Confirming I've set up the IPv6 DNS records for amy01 and ocp13. 

I am unsure if the netplan configuration is something we want to add in salt?
If / when we re-deploying this registry service to future hardware we will probably have different IPs to configure anyway.
For the time being, I have documented the above actions in our server build guide. 

Other than the above question, the items in this issue are resolved.

myroslav:
@RobHooper, IP address will also be part of spoonbill/flatten app config. It can be either ENV variable or entry in the .env file. So the IP is server dependent, but we cannot avoid it in neither in server's network config nor in the app config.

jpmckinney:
@RobHooper Where is the server build guide?

[...]

Issue #145, opened 2020-05-11
Upgrade servers to Ubuntu 20.04


jpmckinney:
~~Once available on Bytemark.~~ Document the process.

~~Can consider moving some~~Move servers to Linode, if upgrading involves provisioning a new server. ~~(Bytemark has very responsive support; I haven't had to interact with Linode recently.)~~

Then, change the [Ubuntu version](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) in GitHub Actions workflows:

* [x] cove-oc4ids (to 22.04)
* [x] cove-ocds (to 22.04)
* [ ] kingfisher-collect
* [ ] kingfisher-process
* [ ] kingfisher-summarize

jpmckinney:
![](https://github.trello.services/images/mini-trello-icon.png) [Upgrade servers to Ubuntu 20.04](https://trello.com/c/rQaUdh4r/195-upgrade-servers-to-ubuntu-2004)

RobHooper:
Ubuntu 20.04.01 (the point release that marks this as production ready) should be released on 06 Aug 2020. I imagine Bytemark will start supporting it from then. 
https://wiki.ubuntu.com/FocalFossa/ReleaseSchedule

jpmckinney:
Noting some changes we'll likely have to make:

https://github.com/OpenDataServices/opendataservices-deploy/commit/ebf4949d3386fab5131dc6ab7dfdfccb34ef02b5

Bjwebb:
BTW, there's still no option to start a server with Ubuntu 20.04 on Bytemark. We've been testing it out by creating a 18.04 server and upgrading it.

RobHooper:
I reached out to Bytemark support and they have no plans to add Ubuntu 20.04 to their list. :frowning_face: 
They also are not planning on removing the Ubuntu 14.04 which has been end of life for months.

We should be able to workaround this using their custom ISO option.

jpmckinney:
@RobHooper Are there pros/cons to (incrementally) moving to a similar host that does offer Ubuntu 20.04?

We use a Linode VPS for the CRM, and we have a Hetzner account for dedicated servers, which we can also use for their Cloud offering. I don't know how the hardware/performance and support compare.

RobHooper:
It's definitely worth looking at other providers and seeing how they compare.
We've been very happy with Linode - there are lots of hosting options out there so Ubuntu 20.04 sounds like a good time to review our options.

[...]

Issue #309, opened 2021-09-24
Add develop/update/docker.rst file


jpmckinney:
Some notes. See `python.rst` as a good example to follow.

Should also include a section on `docker_apps`.

* [x] https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file
* [x] Use of `docker-compose` vs `docker stack`: https://vsupalov.com/difference-docker-compose-and-docker-stack/

## ✔ Browsing

```
docker system df
docker system info
docker container ls -a
docker image ls -a
docker network ls
docker volume ls
```

## ✔ Volumes

To see a volume's data:

```
ls /var/lib/docker/volumes/{volume}/_data/*
```

Volumes are used by Spoonbill for:
- Redis
- PostgreSQL

## ✔ Pruning

https://docs.docker.com/config/pruning/

```
docker volume ls --filter dangling=true
docker image ls --filter dangling=true
docker system prune
```

Note that `docker system prune -a` will remove the containers for cbom and migrate, which are run once on `up` – but maybe these shouldn't be containers to begin with, see #305.

We should see what `docker system prune -a` deletes next time. For reference, minimal images are currently:

```
Minimal images:
REPOSITORY                                                   TAG           IMAGE ID       CREATED       SIZE
docker.pkg.github.com/open-contracting/spoonbill-web/front   sha-80102db   f7ac9167c12a   2 weeks ago   139MB
docker.pkg.github.com/open-contracting/spoonbill-web/app     sha-fb3cd1d   b10151d382ac   2 weeks ago   220MB
ghcr.io/open-contracting/data-registry-static                latest        c902c1bede14   3 weeks ago   33.1MB
ghcr.io/open-contracting/data-registry-django                latest        a58de19373ca   3 weeks ago   982MB
ghcr.io/open-contracting/kingfisher-process                  latest        602f813928bb   3 weeks ago   1.03GB
ghcr.io/open-contracting/pelican-frontend-vue                latest        c0a9edb759c8   6 weeks ago   38.1MB
ghcr.io/open-contracting/pelican-frontend-django             latest        2178f0c65ea8   6 weeks ago   1.05GB
ghcr.io/open-contracting/kingfisher-collect                  latest        0c9bef737fe5   6 weeks ago   968MB
ghcr.io/open-contracting/pelican-backend                     latest        7756121bd86a   6 weeks ago   989MB
traefik                                                      v2.4          77d3b9e70c48   6 weeks ago   92MB
[...]

PR #316, opened 2021-09-30
Ubuntu20 redash

RobHooper:
This PR updates salt to deploy redash on the new ocp14 server running ubuntu 20.04 #145

I have also taken this opportunity to move postgres out of docker and upgrade it #299. Because of this change and specifically [`vm.overcommit_memory`](https://github.com/open-contracting/deploy/blob/main/salt/postgres/init.sls#L82-L84), the memory requirements on this server has increased to needing at least 6GB SWAP / 4GB RAM (from 2GB SWAP / 4GB RAM)

Few tasks are remaining before this is merged in:
- [ ] Update Documentation for onboarding.sls change
- [ ] Import the redash database and test the site
- [ ] Configure networking for the new IPv6 linode block #278
- [ ] Update the redash set up docs https://ocdsdeploy.readthedocs.io/en/latest/deploy/redash.html

@jpmckinney  / @yolile , I can see redash is configured to send email through SES, what is this used for? If I import the database from ocp08 for testing will it trigger any emails?


salt/top.sls
```diff
@@ -48,6 +48,7 @@ base:
 
   'redash':
     - redash
+    - postgres
```
jpmckinney, line 51: I figure this should be first?

salt/redash/init.sls
```diff
@@ -1,16 +1,24 @@
...
     - makedirs: True
+    - template: jinja
+
+/opt/redash/docker-compose.yml:
+  file.managed:
+    - source: salt://redash/files/docker-compose.yml
+    - mode: 644
+    - template: jinja
+
+{{ create_pg_database( salt['pillar.get']('redash:postgres:database'), salt['pillar.get']('redash:postgres:username') ) }}
```
jpmckinney, line 24: It's unusual to use `pillar.get` here. If the configurations are optional, then we should put an if-statement around the call to the `create_pg_database` macro – otherwise, we're running the macro with null values.

salt/redash/init.sls
```diff
@@ -1,16 +1,24 @@
...
 
-redash requirements:
-  pkg.installed:
-    - pkgs:
-      - pwgen
-      - postgresql-client-10
-
-/opt/redash:
-  file.directory:
+# Create Redash configuration and directories
```
jpmckinney, line 10: Is it possible/appropriate to user `docker_apps` like in `registry/init.sls` ? If so, then the env file would instead be configured mostly in Pillar directly.

I haven't yet written up the docs page for it. #309

salt/redash/init.sls
```diff
@@ -1,16 +1,24 @@
+{% from 'lib.sls' import create_pg_database %}
+
 include:
   - docker
 
 # =================================================================================================
 # See https://ocdsdeploy.readthedocs.io/en/latest/deploy/redash.html for installation instructions.
 # =================================================================================================
```
jpmckinney, line 9: I think this can now be removed? (And the linked docs updated.)

salt/redash/files/env
```diff
@@ -0,0 +1,14 @@
...
+REDASH_SECRET_KEY={{ salt['pillar.get']('redash:secret_key') }}
+REDASH_DATABASE_URL=postgresql://{{ salt['pillar.get']('redash:postgres:username') }}:{{ salt['pillar.get']('redash:postgres:password') }}@{{ salt['pillar.get']('redash:postgres:hostname') }}/{{ salt['pillar.get']('redash:postgres:database') }}
+REDASH_FEATURE_SHOW_PERMISSIONS_CONTROL=true
+REDASH_MAIL_SERVER={{ salt['pillar.get']('redash:mail:server') }}
+REDASH_MAIL_PORT={{ salt['pillar.get']('redash:mail:port') }}
+REDASH_MAIL_USE_TLS=true
+REDASH_MAIL_USERNAME={{ salt['pillar.get']('redash:mail:username') }}
+REDASH_MAIL_PASSWORD={{ salt['pillar.get']('redash:mail:password') }}
+REDASH_MAIL_DEFAULT_SENDER={{ salt['pillar.get']('redash:mail:default_sender') }}
+REDASH_HOST={{ salt['pillar.get']('redash:host') }}
```
jpmckinney, line 14: Use `pillar.key1.key2` instead of `salt['pillar.get']('key1:key2')`, unless the value is truly optional.

See https://ocdsdeploy.readthedocs.io/en/latest/develop/style/language.html#jinja-templating-language

salt/redash/files/env
```diff
@@ -0,0 +1,14 @@
+PYTHONUNBUFFERED=0
+REDASH_LOG_LEVEL=INFO
+REDASH_REDIS_URL=redis://redis:6379/0
```
jpmckinney, line 3: Would it be relevant to pull Redis out of Docker, like PostgreSQL? Just wondering. I'm fine to leave it in.

salt/redash/files/env
```diff
@@ -0,0 +1,14 @@
...
+REDASH_COOKIE_SECRET={{ salt['pillar.get']('redash:cookie_secret') }}
+REDASH_SECRET_KEY={{ salt['pillar.get']('redash:secret_key') }}
+REDASH_DATABASE_URL=postgresql://{{ salt['pillar.get']('redash:postgres:username') }}:{{ salt['pillar.get']('redash:postgres:password') }}@{{ salt['pillar.get']('redash:postgres:hostname') }}/{{ salt['pillar.get']('redash:postgres:database') }}
```
jpmckinney, line 6: The current server also sets `POSTGRES_PASSWORD` but not sure if it's actually used by Redash.

salt/postgres/files/conf/redash.conf
```diff
@@ -0,0 +1,5 @@
+{% include 'postgres/files/conf/shared.include' %}
+
+data_directory = '/var/lib/postgresql/13/main'
+
+shared_buffers = 128MB
```
jpmckinney, line 5: Why this value?

salt/onboarding.sls
```diff
@@ -10,16 +10,14 @@ update all packages:
     - refresh: True
     - dist_upgrade: True
 
-{%- if 'fqdn_ip4' in grains %}
-{{ grains.fqdn_ip4[0] }}:
+{{ pillar.main_ipv4 }}:
```
jpmckinney, line 13: This means adding IPs for all servers in Pillar?

salt-config/roster
```diff
@@ -8,11 +8,13 @@ docs:                ocp07.open-contracting.org
...
-redash:              ocp08.open-contracting.org
+redash:              ocp14.open-contracting.org
 registry:
   host: ocp13.open-contracting.org
   port: 2223
 spoonbill-dev:       ocp12.open-contracting.org
 toucan:              ocp09.open-contracting.org
 
+# ocp00 is the Redmine CRM server
+# ocp08 was redash on Ubuntu 18
```
jpmckinney, line 19:
```suggestion
# ocp0 is crm.open-contracting.org
# ocp08 was redash.open-contracting.org on Ubuntu 18
```

pillar/redash.sls
```diff
@@ -9,5 +9,26 @@ apache:
...
-    # 1.22.0 is used in https://github.com/getredash/setup/blob/cb47626b6823cbafac407b3e8991e97f53121f6e/setup.sh#L18
-    version: 1.22.0
\ No newline at end of file
+    version: 1.29.2
+
+redash:
+  docker_tag: redash/redash:10.0.0-beta.b49597
+  postgres:
+    username: redash_user
+    hostname: ocp14.open-contracting.org
```
jpmckinney, line 18: Why can't it connect on localhost?

Some docker-compose files use `network_mode: "host"`, but I know this has its own disadvantages.

jpmckinney (commented):
For the docker-compose file, the differences are:

* remove postgres
* upgrade redis and add healthcheck
* change docker-compose file version
* change docker image tag

salt/redash/init.sls
RobHooper: Updated to meet pillar.get standards matching docs ✔

salt/top.sls
RobHooper: Doesn't matter for the redash application but swapped order so it is alphabetical.

salt/redash/init.sls
RobHooper: I have updated redash to use `docker_apps`.

salt/redash/files/env
RobHooper: This variable is not used AFAICS.

salt/onboarding.sls
RobHooper: I have changed how this works, refer to new server onboarding and netplan docs.

RobHooper:
@jpmckinney , this PR is ready to review. 

I have...
* Updated redash to use `docker_apps`
* Added netplan configuration enabling us to configure IPv6 on Linode servers
* With the netplan changes, I have added configs to registry addressing issue #278 
* Updated documentation around onboarding and created docs for netplan
* Updated redash version for the new stable release
* Updated redash documentation for new setup and processes.

I have also imported the database from the old server onto ocp14 and tested, confirming the app is working fine :)

RobHooper:
Noting migration steps:

1. Merge PR and changes in private repo
    - https://github.com/open-contracting/deploy-pillar-private/pull/24
2. Follow Documented ["Initialize Redash" process](https://github.com/open-contracting/deploy/blob/c4038b9f1f83616544a1182657410e5432f04f00/docs/deploy/redash.rst#initialize-redash)
3. Copy over database from OCP with one additional step:
```bash
sed -i 's/OWNER TO postgres/OWNER TO redash_user/g; /COMMENT ON DATABASE postgres/d' redash.sql
```
4. Update DNS

salt/redash.sls
```diff
@@ -0,0 +1,6 @@
+{% from 'lib.sls' import create_pg_database %}
+
+include:
+  - docker_apps
+
+{{ create_pg_database( "redash_db", "redash_user" ) }}
```
RobHooper, line 6: @jpmckinney I see the `create_pg_database` macro has been removed, how is this now configured?

salt/redash/init.sls
RobHooper: Header removed.

salt/redash.sls
RobHooper: I've modified this to use the new system in `salt/postgres/init.sls` / putting the databases in pillar.
Might need some tweaking, I couldn't find any documentation or use of this new set up yet.

jpmckinney: It's step 2 here. I'll fix the heading for the section to make it clearer.

jpmckinney: https://ocdsdeploy.readthedocs.io/en/latest/develop/update/postgres.html#add-service-accounts

salt/core/network/init.sls
```diff
@@ -0,0 +1,61 @@
...
+      - {{ pillar.host_id }}.open-contracting.org
+      - {{ pillar.host_id }}
+{% endif %}
+
+/etc/mailname:
+  file.managed:
+    - order: 5
+    - contents: "{{ pillar.host_id }}.open-contracting.org"
+
+# The Salt system.networking state does not fully support Ubuntu 20.04 yet so we are using cmd.run instead.
```
jpmckinney, line 26: Can you provide a link to an issue?

pillar/redash.sls
```diff
@@ -1,3 +1,19 @@
...
+
+network:
+  netplan: True
+  ipv4:
+    primary_ip: 139.162.199.85
+    primary_ip_subnet_mask: "/24"
+    gateway_ip: 139.162.199.1
+    dns_servers: [ 178.79.182.5, 176.58.107.5, 176.58.116.5, 176.58.121.5, 151.236.220.5, 212.71.252.5, 212.71.253.5, 109.74.192.20, 109.74.193.20, 109.74.194.20 ]
+  ipv6:
+    primary_ip: 2a01:7e00:e000:02cc::14
```
jpmckinney, line 11: Where is this from? I don't see any IP ending in `::14` at https://cloud.linode.com/linodes/30254827/networking

Also, why don't we need to set a `primary_ip_subnet_mask`? It defaults to 128, but I thought our mask was 64:

```
Open a support ticket with Linode to add the new server to our ``2a01:7e00:e000:02cc::/64`` IPv6 range.
```

docs/deploy/create_server.rst
```diff
@@ -16,35 +16,48 @@ Linode
...
+      #. Click *Edit* for the "My Ubuntu 20.04 LTS Disk Profile" (or similar) config
+      #. Uncheck *Auto-configure networking*
+      #. Click *Save Changes*
 
-#. Power On Linode
+   #. Click *Power On*
+   #. Copy *SSH Access* to your clipboard
 
-#. Select your new server and copy the *SSH Access* details
+#. Open a support ticket with Linode to add the new server to our ``2a01:7e00:e000:02cc::/64`` IPv6 range.
```
jpmckinney, line 56: Where/how is this IPv6 range requested/managed?

pillar/registry.sls
```diff
@@ -1,3 +1,29 @@
+network:
+  host_id: ocp13
```
jpmckinney, line 2: Right now mailname is amy01.open-contracting.org and host files have both ocp13 and amy01. Do we want to change it with this commit?

docs/develop/update/network.rst
```diff
@@ -0,0 +1,85 @@
...
+To fill in the above, from the *Network* tab on the `Linode's <https://cloud.linode.com/linodes>`__ page, collect:
+
+``ipv4``
+  The *Address* with a *Type* of *IPv4 – Public*
+``ipv6``
+  TODO
+``ipv4_subnet_mask``
+  TODO
+``ipv6_subnet_mask``
+  TODO
```
jpmckinney, line 60: @RobHooper Can you fill in the TODOs I added here? I don't know where to find these values from Linode.

docs/develop/update/network.rst
```diff
@@ -0,0 +1,89 @@
...
+``addresses``
+  The *Address* with a *Type* of *IPv6 – SLAAC*, suffixed by "/128"
+``gateway4``
+  The *Default Gateway* with a *Type* of *IPv4 – Public*
+``gateway6``
+  The *Default Gateway* with a *Type* of *IPv6 – SLAAC*
+``nameservers.addresses``
+  The *DNS Resolvers*
+``nameservers.search``
+  Default ``[ open-contracting.org ]``
```
jpmckinney, line 72: Please document in what case the default would be overridden.

pillar/redash.sls
```diff
@@ -1,3 +1,17 @@
+network:
+  host_id: ocp14
+  ipv4: 139.162.199.85
+  ipv6: 2a01:7e00:e000:02cc::14
+  netplan:
+    configuration: linode
+    ipv4_subnet_mask: "/24"
+    addresses:
+      - 2a01:7e00::f03c:92ff:fea5:0e5f/128  # SLAAC
```
jpmckinney, line 9: Why is /128 added to the SLAAC IP? I only see the IP (not the mask) on Linode.

pillar/redash.sls
```diff
@@ -6,7 +20,34 @@ apache:
...
+  version: 13
+  # Redash connections to Postgres are not sent over the localhost connection.
+  # Postgres access is locked down further in the Linode firewall.
+  public_access: true
+  configuration: redash
+  storage: ssd
+  type: oltp
+  nr_hugepages: 128
+  databases:
+    redash_db:
```
jpmckinney, line 34: FYI, instead of having every database be named `X_db`, `Y_db`, etc., we just name database `X`, `Y`, etc.

docs/deploy/redash.rst
```diff
@@ -66,110 +47,86 @@ Run script
...
 
-         docker-compose up -d postgres
+#. Start the Redash application:
 
-   #. Load the database dump using the PostgreSQL credentials:
+   .. code-block:: bash
 
-      .. code-block:: bash
+      cd /data/deploy/redash/
+      docker-compose up -d
```
jpmckinney, line 81: We're starting the containers as root? Shouldn't we have a non-root user?

docs/deploy/create_server.rst
```diff
@@ -190,11 +203,13 @@ Hostnames follow the format ``ocp##.open-contracting.org`` (ocp01, ocp02, etc.).
...
 
-#. Run the `onboarding <https://github.com/open-contracting/deploy/blob/main/salt/onboarding.sls>`__ and core state files, which upgrade all packages, configure the hostname, and apply the base configuration. Replace ``TARGET`` and ``ocpXX``:
+#. :doc:`../develop/update/network`.
+
+#. Run the `onboarding <https://github.com/open-contracting/deploy/blob/main/salt/onboarding.sls>`__ and core state files, which upgrade all packages, configure the hostname and apply the base configuration.
 
    .. code-block:: bash
 
-      salt-ssh --log-level=trace TARGET state.apply 'onboarding,core*' pillar='{"host_id": "ocpXX"}'
+      salt-ssh --log-level=trace TARGET state.apply 'onboarding,core*'
```
jpmckinney, line 212: Should the create_server process also link to https://ocdsdeploy.readthedocs.io/en/latest/develop/update/firewall.html?

pillar/redash.sls
```diff
@@ -6,7 +20,34 @@ apache:
       servername: redash.open-contracting.org
       context:
         proxypass: http://localhost:9090/
+
+postgres:
+  version: 13
+  # Redash connections to Postgres are not sent over the localhost connection.
+  # Postgres access is locked down further in the Linode firewall.
```
jpmckinney, line 27: Does this involve more than what's currently documented at https://ocdsdeploy.readthedocs.io/en/latest/develop/update/firewall.html#linode-vps-servers ?

pillar/redash.sls
```diff
@@ -6,7 +20,34 @@ apache:
       servername: redash.open-contracting.org
       context:
         proxypass: http://localhost:9090/
+
+postgres:
+  version: 13
+  # Redash connections to Postgres are not sent over the localhost connection.
```
jpmckinney, line 26: Can you remind me why this is?

salt/core/network/init.sls
jpmckinney: Aha, https://github.com/open-contracting/deploy/issues/278#issuecomment-924485063.

pillar/redash.sls
jpmckinney: If I remember correctly, it's because Docker uses the external IP.

docs/deploy/redash.rst
jpmckinney: Aha, we do have a `deployer` user, so I'll add `su deployer`.

salt/docker_apps/files/redash.yaml
```diff
@@ -0,0 +1,57 @@
+# Based on: https://github.com/getnewdash/setup/blob/master/data/docker-compose.yml
+version: "2.1"
+x-redash-service: &redash-service
+  image: redash/redash:10.0.0.b50363
```
jpmckinney, line 4: Should we use 10.1.0.b50633 https://hub.docker.com/r/redash/redash/tags ?

docs/deploy/create_server.rst
RobHooper: I created a ticket with Linode asking them to provision a /64 block to us.
https://cloud.linode.com/support/tickets/16265355

We need a /64 block because this is the resolution SPAM blocklists work on. Having the whole range this protects us from "noisy" neighbours sending spam.

At the moment IPv6 management in Linode is entirely done by their support team via support tickets. 

I'll add a link to the Linode support portal to these docs so it is to hand.
https://cloud.linode.com/support/tickets

docs/deploy/create_server.rst
RobHooper: Yes, I've just added a "Firewall" step linking to these docs. 
It should only be needed for docker deployments.

pillar/redash.sls
RobHooper: I think I assumed it was a single IP address which should be indicated with `/128`.
Netplan requires a subnet_mask to be defined.

Looking again I see the Linode Networking page shows a subnet mask of `ffff:ffff:ffff:ffff::` which is `/64`.
This is also in the Linode documentation: https://www.linode.com/docs/guides/linux-static-ip-configuration/#ubuntu-1804-and-2004

Both have worked in my testing - we are not using SLAAC since we are using static configuration but we have to set it up for netplan.

RobHooper: I'll update this to /64.

pillar/redash.sls
RobHooper: We have all IP addresses under `2a01:7e00:e000:02cc:...` and this is for our entire Linode network - shared across all severs.
I chose `::14` because this is running on OCP14.

I only expect to only require one IPv6 address (`/128`) per server most of the time.

pillar/redash.sls
RobHooper: The ocp14 Linode firewall has a allow line granting Postgres access to it's own IPs.
I'm not sure where it would be best to document this? perhaps on the [redash page?](https://ocdsdeploy.readthedocs.io/en/latest/deploy/redash.html#create-a-redash-server)
This is a docker quirk mainly.

pillar/redash.sls
jpmckinney: Aha, we can update the docs to explain how to assign IP addresses without duplication.

pillar/redash.sls
RobHooper: Reviewing the firewall we removed the requirement for the complex "allow self" line now we've got host.docker.local working as the database connection. So the firewall is much simpler. 
I've added this to the redash page as well.

pillar/redash.sls
RobHooper: Correct - Docker is connecting to Postgres in a strange way.

Containers are running in their own network on the `172.17.0.0/16` range (at least by default).
The Redash containers are not aware of any network outside of itself, additionally "localhost" is the container not the server.

salt/redash.sls
RobHooper: Happy the PR is updated for new postgres install + configuration process.

docs/develop/update/network.rst
RobHooper: I've simplified the config, removing some of the values we're not likely going to change (subnet_mask, nameservers and search).
I've filled in the TODO's for the remaining items.

docs/develop/update/network.rst
RobHooper: I can't see a time when these would be overwritten by us.
I've removed these as configurable pillar items simplifying the config.

pillar/redash.sls
```diff
@@ -29,6 +29,8 @@ postgres:
   # Docker containers don't use localhost to connect to the host's PostgreSQL service. Public access is controlled using Linode's firewall.
   public_access: True
   configuration: redash
+  storage: ssd
+  type: oltp
```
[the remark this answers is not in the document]
RobHooper: True, I'm happy either way. We can remove the explicit definition if you prefer.

pillar/registry.sls
RobHooper: Agreed, let's make this change on it's own PR.
This PR sets the groundwork for #278 but doesn't have to implement it.

salt/core/network/files/netplan_linode.yaml
```diff
@@ -0,0 +1,23 @@
...
+  version: 2
+  renderer: networkd
+  ethernets:
+    eth0:
+      {#- Disable auto configuration so we can statically compile. #}
+      dhcp4: no
+      dhcp6: no
+      accept-ra: no
+      addresses:
+	{#- networkd makes the latest IP the primary IP so order matters here #}
```
jpmckinney: I assume "latest" means "last"? I changed it to avoid ambiguity.

That said, https://www.linode.com/docs/guides/linux-static-ip-configuration/#ubuntu-1804-and-2004 puts the SLAAC address last. It also includes a private IP. I don't know if "IPv6 – Link Local" in Linode matches that private IP or is something else.

jpmckinney (commented):
Almost there!

docs/deploy/create_server.rst
```diff
@@ -16,35 +16,50 @@ Linode
...
 
-#. Select your new server and copy the *SSH Access* details
+      #. Click *Edit* for the "My Ubuntu 20.04 LTS Disk Profile" (or similar) config
+      #. Uncheck *Auto-configure networking*
+      #. Click *Save Changes*
+
+   #. Click *Power On*
+   #. Copy *SSH Access* to your clipboard
+
+#. Open a `support ticket with Linode<https://cloud.linode.com/support/tickets>`__ to add the new server to our ``2a01:7e00:e000:02cc::/64`` IPv6 range.
```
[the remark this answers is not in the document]
RobHooper: We need to raise a Linode support ticket for all IPv6 configuration at the ISP level because they don't provide access to configure IPv6 in their support interface. I assume they need to configure this by hand in their routers + switches.
The only thing we need Linode to action is assigning the IPv6 IP to the new server (from our IPv6 range)
Support ticket template:
> Hello,
> Please can you assign our IPv6 IP range (2a01:7e00:e000:02cc::/64) to our new server ocpXX.open-contracting.org.
> Thank you,

In within the next 24 hours I expect a reply just confirming that they've actioned this and we can proceed / re-run just the network set up portion of salt. :)

salt/core/network/files/netplan_linode.yaml
RobHooper: > I assume "latest" means "last"? I changed it
✔✔

In my testing I've had *lots* of issues due to the ordering the IPv6 IPs. The config in this PR is the most reliable set up. 
As I remember, putting SLAAC last will prioritise the other IPv6 address (from the Linode pool) over our IPv6 pool.
The server picks up the Link Local address automatically so no need for manual configuration.

docs/deploy/create_server.rst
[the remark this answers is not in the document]
jpmckinney: Ok, so each time we’re (1) requesting from Linode that the entire range be assigned to the new server - and then (2) in netplan choosing the specific IP (without any involvement from Linode)?

salt/core/network/files/netplan_linode.yaml
jpmckinney: I'm having trouble finding documentation about the priority order. This thread suggests that the first entry is prioritized, not the last: https://bugs.launchpad.net/netplan/+bug/1766656

The manpage for systemd-network doesn't seem to describe any ordering logic: https://manpages.ubuntu.com/manpages/bionic/man5/systemd.network.5.html

jpmckinney: Anyhow, the logic seems to be undocumented/unspecified, so if this config works, let's go ahead with the merge. I've added comments to the file, in case we later discover documented logic.

RobHooper: Agreed - the documentation around Netplan is poor.
I'm happy that this configuration is working fine on OCP14.
I will merge this PR tomorrow morning (20 Apr 2022). :)

PR #324, opened 2021-10-12
registry: add cron for cbom

yolile:


jpmckinney:
I added a commit, since I noticed all `cron.present` IDs weren't setting requirements.

jpmckinney:
Now deployed.

Issue #307, opened 2021-09-24
Follow Docker best practices in production


jpmckinney:
Read and consider steps from:

* https://docs.docker.com/compose/production/
* https://docs.docker.com/engine/install/linux-postinstall/
* https://docs.docker.com/engine/security/rootless/

From call with @yolile:

* [x] Remove volumes from services that don't need/use it
* [x] Change `network_mode`. The Dockerfile should use regular ports, and then docker-compose can expose them appropriately. (Currently, docker-compose passes a port to gunicorn_runner.sh.)

----

Steps I don't think we need to take:

* Check whether to change `restart: unless-stopped` to `restart: always`
  * `unless-stopped` is fine. If an admin manually stops a container, it would be annoying for it to restart always.
* Consider whether to use Docker secrets https://docs.docker.com/engine/swarm/secrets/
  * Requires use of Docker swarm, which is too much trouble.

Other notes:

* Salt should not run docker-compose. There are many components, not all of which need to be updated on each deploy. Also, if done manually, an administrator can ensure that workers finish or shutdown nicely.

Note that I already configured log rotation (i.e. setting `log-driver` to `local`).

yolile:
For containers that need access to host services (eg Postgres databases, RabbitMQ, etc) use the same approach as in  https://github.com/open-contracting/deploy/commit/4cf8abb7b450f021eb2ab746e9e002ac9180c111

jpmckinney:
Closing. If we find any other best practices to implement when reviewing docs in #309, we can create new issues.

PR #327, opened 2021-10-28
Docker updates

yolile:
To be merged after:
- https://github.com/open-contracting/data-registry/pull/170
- https://github.com/open-contracting/pelican-backend/pull/44
- https://github.com/open-contracting/pelican-frontend/pull/68
- https://github.com/open-contracting/kingfisher-process/pull/344


salt/docker_apps/files/pelican_backend.yaml
```diff
@@ -2,11 +2,6 @@ version: "2.2"
 
 x-shared: &shared
   image: "ghcr.io/open-contracting/pelican-backend:latest"
-  network_mode: "host"
```
jpmckinney, line 5: Pelican backend needs to communicate with PostgreSQL and RabbitMQ on the host, and it needs to make remote connections. Can it still do those without `network_mode: "host"`?

jpmckinney (commented):
Also need to update https://github.com/open-contracting/deploy/blob/main/salt/docker_apps/init.sls#L19-L21

salt/docker_apps/files/kingfisher_process.yaml
```diff
@@ -3,8 +3,6 @@ version: "2.2"
 x-shared: &shared
   image: "ghcr.io/open-contracting/kingfisher-process:latest"
   network_mode: "host"
-  env_file:
-    - env
   volumes:
     - /data:/data
```
jpmckinney, line 7: Should this also be /data/storage? (I haven't checked if it reads outside there.)

yolile: I didn't check for the kingfisher-process Dockerfile yet. From a quick look, I see the /data directory is created in the Dockerfile. From the deploy repo I see https://github.com/open-contracting/deploy/blob/9298e0690da0190dd62221a84d6b2ff0d1c14689/pillar/registry.sls#L54 but we are not setting the `KINGFISHER_API_LOCAL_DIRECTORY` as in https://github.com/open-contracting/deploy/blob/9298e0690da0190dd62221a84d6b2ff0d1c14689/pillar/kingfisher.sls#L78, so I'm not sure how kingfisher-collect and kingfisher-process are connected in the registry server and therefore, for what the /data directory is used in kingfisher-process.

salt/docker_apps/files/pelican_backend.yaml
yolile: Oh, I forgot about that. We could do the same as https://github.com/open-contracting/deploy/pull/316/commits/4cf8abb7b450f021eb2ab746e9e002ac9180c111 to still be able to communicate with the host's Postgres and RabbitMQ services, or we can leave this the network_mose as "host".

salt/docker_apps/files/kingfisher_process.yaml
jpmckinney: `KINGFISHER_API_LOCAL_DIRECTORY` is only required for the old Process integration. It is not used by the new Process integration. In both cases, however, Process reads the JSON files written by Collect.

salt/docker_apps/files/pelican_backend.yaml
jpmckinney: We should have the same approach to `network_mode` across all files. We can either change it now to use the approach in 4cf8abb, or open a separate issue.

salt/docker_apps/files/kingfisher_process.yaml
yolile: So Kingfisher process will need to have access to /data/storage/kingfisher-collect? If so we can update KF to create that same directory in its Dockerfile and then update the volume here or leave it as it is now

jpmckinney: Yes, it needs that access. We might as well update its Dockerfile to make it follow the common template: https://ocp-software-handbook.readthedocs.io/en/latest/docker/index.html#templates

yolile: Great to see that you made a template now!

yolile: done in https://github.com/open-contracting/kingfisher-process/pull/344

salt/docker_apps/files/pelican_backend.yaml
yolile: I will leave the network_mode as it is here, and then we can follow up on the existing issue https://github.com/open-contracting/deploy/issues/307

yolile:
I've updated this and the related PRs to use the defaults directories instead of defining them twice: once in the env list and then again in the volumes declaration, as we agreed in our last call

salt/docker_apps/files/kingfisher_process.yaml
```diff
@@ -3,10 +3,8 @@ version: "2.2"
 x-shared: &shared
   image: "ghcr.io/open-contracting/kingfisher-process:latest"
   network_mode: "host"
-  env_file:
-    - env
   volumes:
-    - /data:/data
+    - {{ pillar.kingfisher_collect.files_store }}:/data
```
jpmckinney, line 7: Why not `pillar.kingfisher_collect.env.FILES_STORE` to avoid creating another Pillar key?

yolile: True, changed it now

jpmckinney (approved):

PR #341, opened 2021-11-25
Registry local bind address

yolile:


jpmckinney:
I added a commit to edit `scrapyd.ini`. I also passed the configuration via a jinja context. In general, it's better if templates don't refer to Pillar directly, to make them easier to reuse, but we haven't done this consistently.

PR #342, opened 2021-11-29
registry: update exporter dir to sync with spoonbill

yolile:
to be merged with https://github.com/open-contracting/data-registry/pull/186


jpmckinney (commented):
Why does the mount point inside the container need to change?

jpmckinney:
I think we should instead add an environment variable to the data-registry, and then the registry can join on that to send it to Spoonbill. Otherwise, we have strong coupling between the server, the data registry and spoonbill.

yolile:
Ok, I guess with that solution we will still have some coupling but at least less than the solution in this PR. Closing this in favor of that

Issue #302, opened 2021-09-19
Update Docker configuration for Spoonbill


jpmckinney:
* [x] ~~[v1](https://github.com/open-contracting/deploy/files/6910648/Deployment.instructions.for.spoonbill.pdf) [v2](https://github.com/open-contracting/deploy/files/7312205/Deployment.instructions.for.spoonbill.-.Quintagroup.pdf)~~ [v3](https://github.com/open-contracting/deploy/files/7382998/Deployment.instructions.for.spoonbill.-.Quintagroup.upd.pdf) Quintagroup documentation
* [x] Document and store a copy of the repository secrets (hardcode any non-secrets)
* [x] Move the Docker Compose files to the deploy repository (or add Salt configuration for the git clone on the server)
* [x] Configure PostgreSQL in Salt (including migrating from postgres container) 14fdf1099255ad2bd827fb9bf91c54933e3df0c6
* [x] Change Spoonbill's ACME email
  * Currently, the email is set to shakh@quintagroup.org but it should be sysadmin@open-contracting.org
https://github.com/open-contracting/deploy/pull/348/files#r767065429
* ~~Configure Redis in Salt~~
* ~~Migrate from redis container~~

Redis is okay to leave in Docker, as it's temporary data anyway.

jpmckinney:
> Document and store a copy of the repository secrets (hardcode any non-secrets)

Looks like I did this as there is nothing in https://github.com/open-contracting/spoonbill-web/settings/secrets/actions

jpmckinney:
I've changed the ACME email. Not sure if changing the email at this point will subscribe the new email address, but at any rate it's unlikely we'll have an expiration issue (and if we do, we will address it at that time).

https://doc.traefik.io/traefik/https/acme/
https://letsencrypt.org/docs/expiration-emails/

jpmckinney:
>  Document and store a copy of the repository secrets (hardcode any non-secrets)

The only secret is VUE_APP_SENTRY_DSN, so that the Sentry DSN can be compiled into the static files. The package is public, so anyone could find the DSN, but it's a few too many steps to worry about a bot doing.

https://github.com/open-contracting/spoonbill-web/settings/secrets/actions

> Move the Docker Compose files to the deploy repository (or add Salt configuration for the git clone on the server)

All except the docker-compose.test.yaml file (used with spoonbill-test) has been deleted from the spoonbill-web repository.

PR #348, opened 2021-12-10
spoonbill: initial docker config

yolile:
Ref #302


pillar/spoonbill.sls
```diff
@@ -0,0 +1,22 @@
...
+    version: 1.29.2
+
+docker_apps:
+  spoonbill:
+    target: spoonbill
+    env:
+      DOMAIN: &DOMAIN flatten.open-contracting.org
+      API_DOMAIN: *DOMAIN
+      TRAEFIK_IP: 65.21.93.141
+      ACME_EMAIL: shakh@quintagroup.org
```
jpmckinney, line 14: Please create an issue to change this email address to sysadmin@open-contracting.org (might require some other change with Let's Encrypt - I haven't looked into it)

pillar/spoonbill.sls
```diff
@@ -0,0 +1,22 @@
...
+      DEBUG: 1
```
jpmckinney, line 15: Really?

pillar/spoonbill.sls
```diff
@@ -0,0 +1,22 @@
...
+      DJANGO_ALLOWED_HOSTS: *DOMAIN
+      ALLOWED_HOSTS: *DOMAIN
```
jpmckinney, line 17: It looks like the code only reads `DJANGO_ALLOWED_HOSTS`.

pillar/spoonbill.sls
```diff
@@ -0,0 +1,22 @@
...
+      DB_HOST: postgres
+      REDIS_HOST: redis
+      CORS_ORIGIN_WHITELIST: https://flatten.open-contracting.org
+      SENTRY_ENVIRONMENT: prod
```
jpmckinney, line 21: This is optional for Sentry: https://docs.sentry.io/clients/node/config/ We can remove it.

salt/docker_apps/files/spoonbill.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - '--providers.docker.exposedByDefault=false'
+      - '--entrypoints.http=true'
+      - '--entrypoints.http.address=:80'
+      - '--entrypoints.http.http.redirections.entrypoint.to=https'
+      - '--entrypoints.http.http.redirections.entrypoint.scheme=https'
+      - '--entrypoints.https=true'
+      - '--entrypoints.https.address=:443'
+      - '--log=true'
+      - '--log.level=INFO'
+      - '--certificatesresolvers.acmec.acme.email=${ACME_EMAIL}'
```
jpmckinney, line 44: You can substitute the value from Pillar here. In future it will be sysadmin@ and never change.

jpmckinney (commented):
I didn't check docker-compose.yaml very carefully.

pillar/spoonbill.sls
yolile: Sorry, I used the same variables and values as the current deployment and forgot to remove this one.

pillar/spoonbill.sls
yolile: https://github.com/open-contracting/deploy/issues/349

pillar/spoonbill.sls
```diff
@@ -0,0 +1,22 @@
+docker:
```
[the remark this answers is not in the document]
yolile: Hmm, but Spoonbill is a separated tool, maybe we should move this in a separated and independent file instead?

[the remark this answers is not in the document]
jpmckinney: The logic for Pillar data is server-based, not component-based. That's because the components need to interoperate, and it's easier to configure that in one file.

For Salt files under `salt/`, on the other hand, the logic is component-based, with any server-specific configuration stored in Pillar.

In short, you generally do not mix-and-match Pillar files, but you should be able to mix-and-match Salt files.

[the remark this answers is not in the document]
jpmckinney: When Dogsbody adds the registry backup server, we might split up `pillar/registry.sls` into `pillar/registry_common.sls` to reduce duplication, since the backup will just run the frontend without the other components, but for now one file is simplest.

PR #371, opened 2022-05-04
Apache to wait for networking

### LINKED ISSUE #286: Registry: Apache wait for networking
I rebooted the Registry server and Apache failed to start on boot.
As you can see in the below log output Apache errors trying to bind to 65.21.93.181, I expect networking isn't completely ready yet on the system. 

> Aug 11 10:16:50 amy01 systemd[1]: Starting The Apache HTTP Server...
> Aug 11 10:16:50 amy01 apachectl[710]: (99)Cannot assign requested address: AH00072: make_sock: could not bind to address 65.21.93.181:80
> Aug 11 10:16:50 amy01 apachectl[710]: no listening sockets available, shutting down
> Aug 11 10:16:50 amy01 apachectl[710]: AH00015: Unable to open logs
> Aug 11 10:16:50 amy01 apachectl[694]: Action 'start' failed.
> Aug 11 10:16:50 amy01 apachectl[694]: The Apache error log may have more information.
> Aug 11 10:16:50 amy01 systemd[1]: apache2.service: Control process exited, code=exited, status=1/FAILURE
> Aug 11 10:16:50 amy01 systemd[1]: apache2.service: Failed with result 'exit-code'.
> Aug 11 10:16:50 amy01 systemd[1]: Failed to start The Apache HTTP Server.

We need something to start Apache at the end of the boot process after networking. I think a simple cron running `@reboot` would work but there may be a better way.

RobHooper:
Fixes #286

I didn't use the `systemd` macro because it seemed a bit overkill for what I needed. This approach uploads `/etc/systemd/system/apache2.service.d/customization.conf` which modifies the existing apache2 systemd config file rather than overwriting it.

### DIFF
--- a/salt/apache/init.sls
+++ b/salt/apache/init.sls
@@ -57,3 +57,15 @@ apache2-utils:
 {{ apache(name, entry) }}
 {% endfor %}
 {% endif %}
+
+# Delay the Apache start up if the server has multiple IP addresses.
+{% if salt['pillar.get']('apache:wait_for_networking') == True %}
+/etc/systemd/system/apache2.service.d/customization.conf:
+  file.managed:
+    - source: salt://core/systemd/files/apache2.conf
+    - makedirs: True
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/systemd/system/apache2.service.d/customization.conf
+{% endif %}
--- a/salt/apache/init.sls
+++ b/salt/apache/init.sls
@@ -58,8 +58,8 @@ apache2-utils:
 {% endfor %}
 {% endif %}
 
+{% if salt['pillar.get']('apache:wait_for_networking') %}
 # Delay the Apache start up if the server has multiple IP addresses.
-{% if salt['pillar.get']('apache:wait_for_networking') == True %}
 /etc/systemd/system/apache2.service.d/customization.conf:
   file.managed:
     - source: salt://core/systemd/files/apache2.conf
--- a/pillar/registry.sls
+++ b/pillar/registry.sls
@@ -39,6 +39,7 @@ apache:
   public_access: True
   ipv4: 65.21.93.181
   ipv6: 2a01:4f9:3b:45ca::2
+  wait_for_networking: True
   sites:
     # Can use Scrapyd's basic authentication instead once 1.3 is released.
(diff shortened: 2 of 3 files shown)


salt/apache/init.sls
```diff
@@ -57,3 +57,15 @@ apache2-utils:
 {{ apache(name, entry) }}
 {% endfor %}
 {% endif %}
+
+# Delay the Apache start up if the server has multiple IP addresses.
+{% if salt['pillar.get']('apache:wait_for_networking') == True %}
```
jpmckinney, line 62: Put the comment inside the `if`, so that it isn't added unless the if-statement runs.

```suggestion
{% if salt['pillar.get']('apache:wait_for_networking') %}
# Delay the Apache start up if the server has multiple IP addresses.
```

salt/apache/init.sls
```diff
@@ -57,3 +57,15 @@ apache2-utils:
...
+/etc/systemd/system/apache2.service.d/customization.conf:
+  file.managed:
+    - source: salt://core/systemd/files/apache2.conf
+    - makedirs: True
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/systemd/system/apache2.service.d/customization.conf
```
jpmckinney, line 70: Would this work?

```suggestion
    - watch_in:
      - module: apache2-reload
```

salt/apache/init.sls
RobHooper: Does this also check it is set to True? or is it just checking the `wait_for_networking` is there?
Functionally I suppose they work the same we're just being more explicit with the `== True`

salt/apache/init.sls
RobHooper: `apache2-reload` doesn't work for this change. 
We need to reload the systemctl daemon to pick up on the new systemd file - we don't actually need to reload Apache.

salt/apache/init.sls
RobHooper: Yes, this works as expected in my testing.

salt/apache/init.sls
jpmckinney: Looks good, then!

jpmckinney (approved):

PR #376, opened 2022-07-22
Making NTP pool configurable

RobHooper:


docs/develop/update/network.rst
```diff
@@ -95,3 +95,20 @@ In the server's Pillar file, set ``network.netplan.template`` to ``custom`` and
...
                  ...
+
+NTP Servers
+-----------
+
+If the server is located outside of the UK you will need to manually configure NTP.
+
+You can do this by editing the server's Pillar file and configuring ``ntp`` to a list of the desired `NTP servers <https://www.pool.ntp.org/zone/europe>`__.
+
+For example to use the NTP servers located in Finland:
```
jpmckinney, line 106: Why choose one country or another? Do you choose the country in which the server is located?

You can clarify that the default is to use NTP servers located in the UK (the first paragraph wasn't clear on why servers within the UK don't need any change).

pillar/common.sls
```diff
@@ -1,4 +1,9 @@
 # Values used on all servers. These may be overwritten.
+ntp:
```
jpmckinney, line 2: https://ocdsdeploy.readthedocs.io/en/latest/develop/style/pillar.html sets a common order for keys, to make it easier to scan Pillar files. I think we can put `ntp` as a new line after `vm`, e.g. "Time: `ntp`". And then update its position in the Pillar files in this PR.

RobHooper: Added NTP to the style guide docs and updated pillar to match.

docs/develop/update/network.rst
RobHooper: Docs updated to clarify the above.

jpmckinney (approved):

jpmckinney:
Looks good - can be merged and deployed!

PR #392, opened 2022-11-18
MySQL and Redmine backups to AWS

### LINKED ISSUE #389: MySQL backups
Old documentation:

> A Dogsbody script at /home/sysadmin-tools/bin/mysql_backup_to_gcloud.sh uploads the Redmine database and files to [Google Cloud](https://console.cloud.google.com/storage/browser?project=ocds-172716&organizationId=1015889055088&folder) under the data@open-contracting.org account. This stores 7 daily backups and 5 weekly backups.

We don't use GCP for anything else, so ideally we would switch to AWS.

Once we do that, we can:

* [ ] Delete the above script from the repository
* [ ] Delete the related GCP project

RobHooper:
- [x] Write Documentation
- [x] Create S3 bucket
- [x] Create IAM access keys

Resolves #389

### DIFF
--- a/salt/aws/init.sls
+++ b/salt/aws/init.sls
@@ -1,0 +1,24 @@
+{% from 'lib.sls' import set_config %}
+
+python3-pip:
+  pkg.installed:
+    - name: python3-pip
+
+awscli:
+  pip.installed:
+    - name: awscli
+    - require:
+      - pkg: python3-pip
+
+/home/sysadmin-tools/aws-settings.local:
+  file.managed:
+    - source: salt://aws/files/aws-settings.local
+    - mode: 640
+    - replace: False
+    - require:
+      - file: /home/sysadmin-tools/bin
+
+{{ set_config("aws-settings.local", "AWS_ACCESS_KEY_ID", pillar.aws.access_key) }}
+{{ set_config("aws-settings.local", "AWS_SECRET_ACCESS_KEY", pillar.aws.secret_key) }}
+{{ set_config("aws-settings.local", "AWS_DEFAULT_REGION", pillar.aws.region) }}
+{{ set_config("aws-settings.local", "AWS_CLI", "/usr/local/bin/aws") }}
--- a/salt/aws/init.sls
+++ b/salt/aws/init.sls
@@ -21,4 +21,3 @@ awscli:
 {{ set_config("aws-settings.local", "AWS_ACCESS_KEY_ID", pillar.aws.access_key) }}
 {{ set_config("aws-settings.local", "AWS_SECRET_ACCESS_KEY", pillar.aws.secret_key) }}
 {{ set_config("aws-settings.local", "AWS_DEFAULT_REGION", pillar.aws.region) }}
-{{ set_config("aws-settings.local", "AWS_CLI", "/usr/local/bin/aws") }}
--- a/salt/files/mysql-backup-to-s3.sh
+++ b/salt/files/mysql-backup-to-s3.sh
@@ -10,6 +10,7 @@
 
 set -euo pipefail
 
+# shellcheck disable=SC1091
 source /home/sysadmin-tools/aws-settings.local
 
 export AWS_ACCESS_KEY_ID
@@ -43,7 +44,7 @@ for DATABASE in "${DATABASES[@]}"; do
     information_schema | performance_schema | sys | innodb | mysql) ;; # Skip system databases
     *)
         TIMESTAMP="$(TZ=UTC date +%Y-%m-%d_%H:%M:%S)"
-        TMPFILE=$(mktemp)
+        TMPFILE="$(mktemp mysql_backup_XXXX.sql.gz)"
         ${MYSQLDUMP} "${MYSQL_ARGUMENT}" --databases "${DATABASE}" | gzip > "${TMPFILE}"
         if zgrep -q "Dump completed on" "${TMPFILE}"; then
             $AWS_CLI s3 cp "${TMPFILE}" "s3://${S3_DATABASE_BACKUP_BUCKET}/${TIMESTAMP}_${DATABASE}.sql.gz" --only-show-errors
--- a/salt/files/mysql-backup-to-s3.sh
+++ b/salt/files/mysql-backup-to-s3.sh
@@ -12,74 +12,46 @@ set -euo pipefail
 
 MYSQL="/usr/bin/mysql"
 MYSQLDUMP="/usr/bin/mysqldump"
+MYSQL_DEFAULTS_FILE="/home/sysadmin-tools/mysql-defaults.cnf"
 
 source /home/sysadmin-tools/aws-settings.local
-
-# For storing the backups before pushing into AWS
-TMPFILE=$(mktemp)
-TMPLOG=$(mktemp)
-
-# If a backup fails we want detailed output
-function echo_debug {
-    echo "$(date +%Y-%m-%dT%H:%M:%S): ${*}" >> "${TMPLOG}"
-}
-
-# Enabling debug mode
-echo_debug "Backup script starting"
 
 export AWS_ACCESS_KEY_ID
 export AWS_SECRET_ACCESS_KEY
 export AWS_DEFAULT_REGION
 
-if [ -r "${DB_DEFAULTS_FILE}" ]; then
-    MYSQL_CMD_LOGIN="--defaults-extra-file=${DB_DEFAULTS_FILE}"
-else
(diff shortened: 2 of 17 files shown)


jpmckinney:
For the AWS setup, I can do the same as we did for our other backups?

RobHooper:
@jpmckinney, I have added new documentation covering the Salt configuration.
Would you like docs detailing out IAM user and S3 bucket creation? if so where is best to document this?

The Lint Shell alert is flagging that it is sourcing a file that doesn't exist (which makes sense because it doesn't in the test environment). I expect we've had this issue previously, do you remember how we worked around this?

jpmckinney:
> Would you like docs detailing out IAM user and S3 bucket creation? if so where is best to document this?

Here's the page for AWS docs: https://ocdsdeploy.readthedocs.io/en/latest/deploy/aws.html

> The Lint Shell alert is flagging that it is sourcing a file that doesn't exist (which makes sense because it doesn't in the test environment). I expect we've had this issue previously, do you remember how we worked around this?

Add `# shellcheck disable=SC1091` on the line before

salt/files/site-backup-to-s3.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+if [ ! -x "${AWS_CLI}" ]; then
+    echo "Error: The aws executable is not installed"
+    exit 7
+fi
+
+for SITE in "${FOLDER2S3BACKUPSRC[@]}"; do
+    SITENAME="${SITE//[^a-zA-Z0-9]/_}"
+    SITENAME="${SITENAME//^___/}"
```
jpmckinney: What does this line do?

salt/files/site-backup-to-s3.sh
```diff
@@ -0,0 +1,45 @@
...
+    TIMESTAMP="$(TZ=UTC date +%Y-%m-%d)"
+    LOCALBACKUPLOCATION="/tmp/${SITENAME}backup_${TIMESTAMP}.tar.gz"
+
+    # To stop the script breaking when reading changing files (e.x. logs)
```
jpmckinney, line 35: I'm not clear. If `tar` fails, we still want the script to continue? Can tar partially fail?

salt/files/site-backup-to-s3.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ ! -x "${AWS_CLI}" ]; then
+    echo "Error: The aws executable is not installed"
+    exit 7
+fi
+
+for SITE in "${FOLDER2S3BACKUPSRC[@]}"; do
+    SITENAME="${SITE//[^a-zA-Z0-9]/_}"
+    SITENAME="${SITENAME//^___/}"
+    TIMESTAMP="$(TZ=UTC date +%Y-%m-%d)"
+    LOCALBACKUPLOCATION="/tmp/${SITENAME}backup_${TIMESTAMP}.tar.gz"
```
jpmckinney: The other file uses `mktemp` for the local file.

salt/files/site-backup-to-s3.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    # To stop the script breaking when reading changing files (e.x. logs)
+    set +e
+    if [ -z "${BACKUPEXCLUDE}" ]; then
+        tar czf "${LOCALBACKUPLOCATION}" "${SITE}" &> /dev/null
+    else
+        tar czf "${LOCALBACKUPLOCATION}" "${SITE}" "${BACKUPEXCLUDE}" &> /dev/null
+    fi
+    set -e
+    ${AWS_CLI} s3 cp "${LOCALBACKUPLOCATION}" "s3://${S3FILEBACKUPBUCKET}/${SITENAME}backup_${TIMESTAMP}.tar.gz" --quiet
```
jpmckinney: The other file uses `--only-show-errors` instead of `--quiet`.

salt/aws/init.sls
```diff
@@ -0,0 +1,24 @@
...
+    - require:
+      - pkg: python3-pip
+
+/home/sysadmin-tools/aws-settings.local:
+  file.managed:
+    - source: salt://aws/files/aws-settings.local
+    - mode: 640
+    - replace: False
+    - require:
+      - file: /home/sysadmin-tools/bin
```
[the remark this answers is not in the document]
RobHooper: Agreed, and added `require` `file: /home/sysadmin-tools/bin`.

Better than this would be requiring `/home/sysadmin-tools/{{ filename }}` however this errors as a recursive requisite because of how salt loads it.

salt/files/mysql-backup-to-s3.sh
```diff
@@ -0,0 +1,57 @@
...
+source /home/sysadmin-tools/aws-settings.local
+
+export AWS_ACCESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION
+
+if [ "${LOGNAME}" != "root" ]; then
+    echo "ERROR: Execution of ${0} stopped as not run by user root!"
+    exit 3
+fi
```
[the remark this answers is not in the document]
RobHooper: You are absolutely right this is unnecessary thanks to the file permissions.
I really like having this extra sanity check before the script is ran. I think we should have it on both scripts, they should only be ran as root and this helps ensure it.

salt/mysql/backup.sls
```diff
@@ -0,0 +1,35 @@
...
+
+{{ set_config("aws-settings.local", "S3DATABASEBACKUPBUCKET", pillar.mysql.backup.location ) }}
+
+/home/sysadmin-tools/bin/mysql-backup-to-s3.sh:
+  file.managed:
+    - source: salt://files/mysql-backup-to-s3.sh
+    - mode: 750
+    - require:
+      - file: /home/sysadmin-tools/bin
+      - sls: aws
```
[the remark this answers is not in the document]
RobHooper: Yes, added.

salt/mysql/backup.sls
```diff
@@ -0,0 +1,35 @@
...
+
+# Default to blank if there is no root password.
+/home/sysadmin-tools/mysql-defaults.cnf:
+  file.managed:
+    - contents: |
+       [client]
+       user = root
+       password = {{ salt['pillar.get']('mysql:users:root:password','') }}
+    - user: root
+    - group: root
```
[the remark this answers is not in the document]
RobHooper: This defaults to the user salt is running as - which should be root.
Lines removed to match other files.

salt/files/mysql-backup-to-s3.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [ ! -x "${AWS_CLI}" ]; then
+    echo "Error: The aws executable is not installed"
+    exit 3
+fi
```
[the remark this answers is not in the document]
RobHooper: I like this check because it double checks the environment matches what we are configuring in salt.
It also double checks the `aws` binary has been installed to the location we expect - this can change depending on how aws-cli was installed.

salt/files/site-backup-to-s3.sh
RobHooper: This line converts the site name into an alphanumeric (plus underscores) string, so it ensures it is valid for AWS.
> SITENAME="${SITE//[^a-zA-Z0-9]/_}"

The next line removes the underscore prefix, tidying up the file name, however it isn't as intended.
I've just replaced this line with these:
> SAFENAME="${SAFENAME/#_/}" # Remove prefix
> SAFENAME="${SAFENAME/%_/}" # Remove suffix

salt/files/site-backup-to-s3.sh
RobHooper: If `tar` fails we certainly want the script to continue. 
For example if we were backing up multiple sites and it errors on the first backup, we still want the other sites to be backed up.

Tar will return an error exit code in a number of OK situations, for example, a file changing or disappearing during the backup. This happens often when the site stores logs or cache files under the backup path. 
The backup is still OK and we want the script to continue.

salt/files/site-backup-to-s3.sh
RobHooper: Fixed

salt/files/site-backup-to-s3.sh
RobHooper: Updated.

salt/files/site-backup-to-s3.sh
jpmckinney: Aha, yeah, the second line didn't look right ;)

jpmckinney (approved):

Issue #387, opened 2022-10-11
Elasticsearch: Use PEM files instead of Keystore


jpmckinney:
In the new version of ReadOnlyREST (1.44), we can now use PEM files directly: https://github.com/beshu-tech/readonlyrest-docs/blob/master/elasticsearch.md#pem-files-instead-of-a-keystore-andor-truststore

This implies deleting:

* [x] pem-to-keystore-wrapper.sh
* [x] pem-to-keystore.sh (probably still need to restart ES though)
* [x] sudoers.d/pem-to-keystore (assuming www-data can be given permission to restart ES via `service` command)
* [x] Related lines in readonlyrest.sls

And changing:

* [x] Related keys in `readonlyrest-docs.yml`
* [x] MDNotifyCmd in `docs.sls`
* [x] Docs in `elasticsearch.rst`

RobHooper:
@jpmckinney , this change has been deployed, ES ROR is now using pem files.
However I have found an issue with this approach, ES cannot load the files directly from Apache due to permissions issues.
I've put a temporary solution in place (copying the cert files to a different path).

We could write a script to automatically copy these files - I am worried this might not work, we had a number of issues with the keystore script.
I think a better solution is to proxy traffic through Apache, removing the SSL configuration from ROR.

jpmckinney:
We can proxy, but I assume the URL/port to access ES will change, so I'll need to update all the sites on the server.

What is the specific permission issue? Can we not change the group/permissions on the original files?

jpmckinney:
Ah, so the files are owned by root and set to 600. The main Apache process runs as root, and then it drops privileges by creating child processes running as www-data. This approach is also what the mod_md maintainers expect of other software: https://github.com/icing/mod_md/issues/84

I haven't checked, but I think ES doesn't take this approach.

> when Apache is restarted

Yes, this seems to be the case based on above issue, so we can't use the same paths/files.

> permissions would reset frequently whenever the cert is renewed

This isn't an issue, since we already have `MDNotifyCmd` at renewal, which can modify the certs.

Can we edit the MDNotifyCmd script to copy the certs and allow only ES to read those copies?

jpmckinney:
We had bugs, but they are solved now: it was because Apache couldn't run the script, and we fixed it by setting up sudoers.

Anyway, sounds good to update the command and leave the proxy as a new issue.

jpmckinney:
[...]

PR #395, opened 2022-12-01
Removing keystore and adding pem SSL support

RobHooper:
Addressing issue #387

When we deploy we also need to action:
1. Remove the following files from the server
>  /opt/pkcs-password
>  /opt/pem-to-keystore.sh
>  /opt/pem-to-keystore-wrapper.sh
>  /etc/sudoers.d/90-pem-to-keystore 
2. Remove ``key_pass`` from pillar/private ``elasticsearch:plugins:readonlyrest:key_pass``


salt/elasticsearch/files/config/readonlyrest-docs.yml
```diff
@@ -2,9 +2,8 @@ readonlyrest:
   # https://github.com/beshu-tech/readonlyrest-docs/blob/master/kibana.md#loading-settings-order-of-precedence
   force_load_from_file: True
   ssl:
-    keystore_file: keystore.jks
-    keystore_pass: {{ pillar.elasticsearch.plugins.readonlyrest.key_pass }}
-    key_pass: {{ pillar.elasticsearch.plugins.readonlyrest.key_pass }}
+    server_certificate_key_file: /etc/apache2/md/domains/standard.open-contracting.org/privkey.pem
```
RobHooper, line 5: I don't think this should be hard-coded, perhaps a variable like: ``elasticsearch:plugins:readonlyrest:privkey``

jpmckinney: Sounds good to me.

salt/elasticsearch/plugins/readonlyrest.sls
```diff
@@ -12,45 +12,27 @@ readonlyrest-install:
...
-      - file: /opt/pkcs-password
-
-/opt/pem-to-keystore-wrapper.sh:
-  file.managed:
-    - name: /opt/pem-to-keystore-wrapper.sh
-    - source: salt://elasticsearch/files/pem-to-keystore-wrapper.sh
+    - name: /opt/restart-elasticsearch.sh
+    - source: salt://elasticsearch/files/restart-elasticsearch.sh
     - user: root
     - group: root
```
jpmckinney, line 20: These are from before, but we can probably remove the user/group root lines.

salt/elasticsearch/plugins/readonlyrest.sls
```diff
@@ -12,45 +12,27 @@ readonlyrest-install:
...
     - mode: 755
```
jpmckinney, line 21: Change to 750 if only run via sudo/root?

salt/elasticsearch/files/config/readonlyrest-docs.yml
RobHooper: Done ✔

salt/elasticsearch/plugins/readonlyrest.sls
RobHooper: Done ✔

RobHooper: This needs to be executable (and not editable) for the ``www-data`` user which 750 will block.

Alternatively we could set the group to ``www-data`` however this is quite specific to the docs server and won't scale as easily if other users need access.

salt/elasticsearch/plugins/readonlyrest.sls
RobHooper: Done ✔

salt/elasticsearch/plugins/readonlyrest.sls
jpmckinney: Ah, right, this file is run by www-data, and the command within this file is sudo 👍

jpmckinney (approved):

Issue #381, opened 2022-08-25
Upgrade servers to Ubuntu 22.04


RobHooper:
Ubuntu 22.04 is available on Linode.
I am not expecting many changes to our Salt code as it is quite similar to 20.04. We will need to update documentation for the new version.

Currently there are bugs in SaltStack preventing deployments to 22.04 servers, these need resolving before we can move forward with upgrades.
This bug in-particular completely breaks our use of pkgrepo, https://github.com/saltstack/salt/issues/59065

This will supersede the 20.04 server upgrades #145

Change the Ubuntu and Python #346 versions in GitHub Actions workflows:

* [ ] kingfisher-collect
* [ ] kingfisher-process
* [ ] kingfisher-summarize

jpmckinney:
There are PRs for our blocking issues:

* https://github.com/saltstack/salt/pull/63157 for https://github.com/saltstack/salt/issues/59065 (pkgrepo)
* https://github.com/saltstack/salt/pull/63184 for https://github.com/saltstack/salt/issues/60003, which was keeping us locked to salt-ssh<3003

According to the [3005 release notes](https://docs.saltproject.io/en/latest/topics/releases/3005.html) (Aug 2022), Ubuntu 22.04 is supported.

Note that Salt is changing to a ["one-dir"](https://docs.saltproject.io/salt/install-guide/en/latest/topics/upgrade-to-onedir.html#what-is-onedir) packaging approach for future releases.

Also, if using macOS, new versions of salt-ssh require Python 3.9 or 3.10.

jpmckinney:
Okay, I succeeded in upgrading Salt in requirements.txt so that I can use it locally on Python 3.10, macOS Ventura.

I don't have an Ubuntu 22.04 server to test on. I was able to deploy to 18.04 and 20.04 servers using PPAs and filerefs.

jpmckinney:
Noting that I referenced a couple issues in specific applications that will be unblocked by this issue.

jpmckinney:
It looks like we already updated relevant documentation.

Since we've been deploying to Ubuntu 22, it seems the above issues are no longer blockers.

I've moved the Ubuntu and Python changes in project workflows to #402.

PR #404, opened 2023-01-18
cove-oc4ids Ubuntu upgrade

RobHooper:
This PR adds code to upgrade cove-oc4ids to Ubuntu 22.04. #381.

The `pkgrepo.managed` bug is still present on 22.04, for this reason, I have re-configured Salt to install Apache directly from Ubuntu.
IIRC we were only using Ondrej for mod_md support, this functionality is now provided in the native Ubuntu version.

I have also taken this opportunity to implement `systemd-networkd` network configuration which has proven more resilient on Linode compared to netplan.

Todo:
- [x] Test uWSGI application
- [x] Documentation
- [ ] Plan go-live process


pillar/cove_oc4ids.sls
```diff
@@ -10,9 +18,9 @@ python_apps:
       url: https://github.com/open-contracting/cove-oc4ids.git
     django:
       env:
-        ALLOWED_HOSTS: .standard.open-contracting.org,.oc4ids.opencontracting.uk0.bigv.io
+        ALLOWED_HOSTS: .infrastructure.open-contracting.org,.cove-oc4ids.open-contracting.org
```
RobHooper, line 21: @jpmckinney, I remember us discussing removing the proxy from the cove servers - which I think will be a good upgrade, simplifying hosting and making server upgrades easier.

What hostnames would you like to use for the cove applications?

RobHooper:
@jpmckinney, the cove application is failing to start on this new server (ocp17), please can you investigate?

I have confirmed that the pip packages are all installed correctly.
The uWSGI configuration looks correct too.
There are no errors logged in `/var/log/uwsgi/app/cove.log`.
It seems to fails to find / start the application.

I was wondering if app changes were needed for the new Python version (3.10) such as pip package updates?

jpmckinney:
> @jpmckinney, the cove application is failing to start on this new server (ocp17), please can you investigate?
> 
> I was wondering if app changes were needed for the new Python version (3.10) such as pip package updates?

Yes, requirements.txt had to be rebuilt. I've done that now.

jpmckinney:
> @jpmckinney, I remember us discussing removing the proxy from the cove servers - which I think will be a good upgrade, simplifying hosting and making server upgrades easier.
>
> What hostnames would you like to use for the cove applications?

I'm thinking review.standard.open-contracting.org and oc4ids-review.standard.open-contracting.org

@yolile What do you think?

yolile:
> I'm thinking review.standard.open-contracting.org and oc4ids-review.standard.open-contracting.org

Maybe review-oc4ids.standard.open-contracting.org instead? So it is read something like "review the oc4ids standard"

jpmckinney:
review-oc4ids.standard.open-contracting.org is also good with me. (Though technical it reads as review Open Contracting for Infrastructure Data **Standards standard**).

pillar/cove_oc4ids.sls
RobHooper: Proceeding with `review-oc4ids.standard.open-contracting.org` as discussed.

salt/uwsgi/files/django.ini
```diff
@@ -1,10 +1,13 @@
 [uwsgi]
 ; https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/
-; strict = true errors due to deb-confname in /usr/share/uwsgi/conf/default.ini
```
RobHooper, line 3: I have enabled `strict` on the new hosting and I am not seeing any errors, the app loads correctly.
We probably need to prevent this being accidentally deployed to any of the older hosting - assume it still errors there.

RobHooper:
You can access the pre-live cove application by pointing your hosts file directly to the new server:
> 176.58.112.127	review-oc4ids.standard.open-contracting.org 

I tested the app is running OK and uploaded a file to check functionality but it may be worth someone who knows the app better to check it is functioning all as expected.

Regarding the go-live process, we will need to update anything referencing the old URL to point directly to the new one.
We should also configure a redirect on https://standard.open-contracting.org/infrastructure/review/.
@jpmckinney, please can I leave updating the old URL with you?

jpmckinney:
> You can access the pre-live cove application by pointing your hosts file directly to the new server:
> 
> > 176.58.112.127	review-oc4ids.standard.open-contracting.org
> 
> I tested the app is running OK and uploaded a file to check functionality but it may be worth someone who knows the app better to check it is functioning all as expected.

@yolile Do you have the time to test? If not, it sounds like it's working, and we can debug afterwards if something comes up.

> Regarding the go-live process, we will need to update anything referencing the old URL to point directly to the new one. We should also configure a redirect on https://standard.open-contracting.org/infrastructure/review/. @jpmckinney, please can I leave updating the old URL with you?

I think I've now made the necessary change in this PR at commit 55423a8

jpmckinney:
@RobHooper I've made a commit to cove-oc4ids to stop rewriting its own URL as /infrastructure/review.

pillar/cove_oc4ids.sls
```diff
@@ -1,3 +1,11 @@
+network:
+  host_id: ocp17
+  ipv4: 176.58.112.127
+  ipv6: "2a01:7e00:e000:04c1::"
```
jpmckinney, line 4: Just noticing I might have mis-documented our IPv6 /64 block. In a couple places in the docs I refer to it as `2a01:7e00:e000:02cc::/64`, but maybe it should only be `2a01:7e00:e000:....` ?

salt/core/network/files/networkd_linode.template
```diff
@@ -0,0 +1,20 @@
...
+Address={{ pillar.network.ipv4 }}/24
+{%- if 'ipv6' in pillar.network %}
+Address={{ pillar.network.ipv6 }}/64
+{%- endif %}
+{%- for address in pillar.network.networkd.get('addresses', []) %}
+Address={{ address }}
+{%- endfor %}
+
+Gateway={{ pillar.network.networkd.gateway4 }}
+Gateway=fe80::1
```
jpmckinney, line 20: In the netplan template, gateway6 is configurable. Why is it not here?

salt/core/network/files/networkd_linode.template
```diff
@@ -0,0 +1,20 @@
...
+IPv6PrivacyExtensions=false
+IPv6AcceptRA=false
+
+Address={{ pillar.network.ipv4 }}/24
+{%- if 'ipv6' in pillar.network %}
+Address={{ pillar.network.ipv6 }}/64
+{%- endif %}
+{%- for address in pillar.network.networkd.get('addresses', []) %}
+Address={{ address }}
+{%- endfor %}
```
jpmckinney, line 17: In the netplan template, there's a comment about the last IP being used as the primary IP, and the ipv6 is listed last. Should we do the same here, or does the same behavior not occur?

salt/python/extensions.sls
```diff
@@ -3,3 +3,10 @@ python c extensions:
     - pkgs:
       - python3-dev
       - build-essential
+
+{% if grains.osmajorrelease|string == '22' %}
```
jpmckinney, line 7: Why do we need `|string` here? (One other occurrence)

salt/python_apps.sls
```diff
@@ -50,6 +53,9 @@ include:
     - cwd: {{ directory }}
     - require:
       - virtualenv: {{ directory }}/.ve
+      {% if grains.osmajorrelease|string == '22' %}
+      - pkg: libffi-dev
+      {% endif %}
```
jpmckinney, line 58: This isn't required by all our Python apps, and cove.sls already includes python.extensions before python_apps.

In extensions.sls, we can roll up the libffi-dev state into the "python c extensions" state.

```suggestion
```

yolile:
> You can access the pre-live cove application by pointing your hosts file directly to the new server:

It returns the Apache default site for me (even after editing the `hosts` file)

yolile:
It works now, and the features are working as expected too

jpmckinney:
@RobHooper Hi Rob, just checking on the progress on this?

Once this is done, it should be quick to do cove-ocds as well.

pillar/cove_oc4ids.sls
RobHooper: This is a bigger issue see discussion on #409

salt/core/network/files/networkd_linode.template
RobHooper: The gateway6 address is always the same on Linode so I removed this function simplifying the configuration.

salt/python/extensions.sls
RobHooper: I was doing this because `grains.osmajorrelease` is an integer and it was erroring in the comparison.
I've just committed switching both values to integers.

salt/python_apps.sls
RobHooper: I saw errors from `pip` without this explicit requirement.
However, I can't replicate this issue now.
I've added this suggestion in the latest commit.

salt/core/network/files/networkd_linode.template
jpmckinney: Aha, should we hardcode it in `netplan_linode.yaml` too?

salt/core/network/files/networkd_linode.template
RobHooper: From my testing the order doesn't seem to matter. I'm not sure how it chooses which IP to use for outbound traffic.

I don't expect this to be a problem unless we wanted the SLAAC IP configured on the server as well as the /64 block. I can't think of a use-case when we would want this so, we can investigate more when needed.

jpmckinney: Ok, sounds good!

jpmckinney (approved):

salt/core/network/files/networkd_linode.template
RobHooper: Done

RobHooper:
I'm happy issue #409 is not a blocker.

Go Live Plan:
- [ ] Create new site DNS record
- [ ] Merge this PR into Main
- [ ] Copy app uploads to new server(?)
- [ ] Update app for new URL
- [ ] Update sites referring to old URL

@jpmckinney Rough go-live plan detailed above. Please can you advise on any missing steps. 

I believe the cove-oc4ids repo is ready and deployed to the new server. Are changes needed elsewhere? 
For instance, I would like the old URL on OCP07 to redirect to the new URL - https://standard.open-contracting.org/infrastructure/review/

jpmckinney:
> I believe the cove-oc4ids repo is ready and deployed to the new server. Are changes needed elsewhere?
> For instance, I would like the old URL on OCP07 to redirect to the new URL - https://standard.open-contracting.org/infrastructure/review/

We just need to deploy the ocds-docs server. The necessary changes are in this PR.

Since the redirect will be set up, we don't need to update the old URL(s) immediately.

I'll create the DNS records.

jpmckinney:
review-oc4ids.standard.open-contracting.org is now CNAME to ocp17.open-contracting.org

salt/prometheus/files/conf-prometheus.yml
```diff
@@ -11,11 +11,11 @@ scrape_configs:
 
   # Node Exporter.
 {%- for job_name, host in {
-    'ocds-cove-live.oc4ids': 'ocp01.open-contracting.org',
+    'ocds-cove-live.oc4ids': 'ocp17.open-contracting.org',
```
RobHooper, line 14: @jpmckinney , please could you suggest a better `job_name` for prometheus?
I would rather the instances reported in as new entities rather than replacing the previous servers - happy to be overruled tho.

RobHooper:
> We just need to deploy the ocds-docs server. The necessary changes are in this PR.
@jpmckinney , I've updated this PR to include the cove3 replacement server too.

You can test the new instance by setting the following in your hosts file:
```
176.58.107.239 review.standard.open-contracting.org
```
I've uploaded a test file and it appears to be working OK so I am happy putting new DNS live, if you are?

salt/prometheus/files/conf-prometheus.yml
jpmckinney: Sure, maybe we just use the same name as in the Salt roster, and increment with each time that service has been redeployed? So it would now be `cove-oc4ids2` (if we pretend the original was `cove-oc4ids1` instead of `ocds-cove-live.oc4ids`).

jpmckinney: Then for the other one: cove-ocds-2 (I don't know where ocds-live3 comes from, and I don't care to find out ;)).

jpmckinney:
I've added the CNAME review.standard.open-contracting.org to ocp18, since it doesn't interfere with anything and makes testing easier.

jpmckinney:
I'm seeing an SSL warning for https://review.standard.open-contracting.org. It's also redirecting to /review/ because I haven't merged https://github.com/open-contracting/cove-ocds/pull/177. I'll merge and re-deploy.

RobHooper:
> I'm seeing an SSL warning for https://review.standard.open-contracting.org/.

I've just fixed the SSL warning, I had a temporary solution in place while DNS wasn't live.

jpmckinney:
Hmm, well I merged the PR but now I'm getting `jinja2.exceptions.TemplateNotFound: lib.sls` errors when trying to deploy anything.

jpmckinney:
> Copy app uploads to new server

Why is still crossed out? Doesn't it still need to be done after ocds-docs is deployed?

jpmckinney:
> Update app for new URL
> Update sites referring to old URL

I'm not sure what these refer to.

jpmckinney:
> Hmm, well I merged the PR but now I'm getting jinja2.exceptions.TemplateNotFound: lib.sls errors when trying to deploy anything. (Not related to the changes in this PR – Salt is just broken software.)

Ok, I got Salt to work by manually applying the patch https://github.com/saltstack/salt/pull/63184 (needed until salt-ssh==3006).

I deployed cove-ocds, cove-oc4ids, docs and prometheus.

>> Copy app uploads to new server
> Why is still crossed out? Doesn't it still need to be done after ocds-docs is deployed?

I've copied the media directory and the db.sqlite3 file to cove-ocds, [as documented](https://ocdsdeploy.readthedocs.io/en/latest/deploy/create_server.html).

There were only 5 reports in the old cove-oc4ids, so I didn't bother copying them over to the new server.

jpmckinney:
I've added decommissioning the old servers as a step to the go-live plan.

RobHooper:
Awesome this is coming together well :+1:

Well done finding the bug in salt. I haven't been affected by this yet, probably quirks from different software versions on my system.

> Update app for new URL
This is to update the cove application for the new URL.

You've actioned this, I believe these are the corresponding commits:
https://github.com/open-contracting/cove-ocds/pull/177
https://github.com/open-contracting/cove-oc4ids/commit/315adaa6b6c64698820e30bacfed596caf29bb23

> Update sites referring to old URL
This is to update anything that points to the cove review applications.

For instance the "Data Review Tool" button on the standard site: https://standard.open-contracting.org/latest/en/
I am not sure how many links are around still configured for the old URL.

Since we have redirects configured, this isn't a problem but it would be good to fix them as we find them. 
Extra, unnecessary, redirects will slow down site visitors - adding an extra request whenever they visit the page.

> Copy app uploads to new server
Thanks for sorting app media, I had the impression (incorrectly) that this step wasn't needed but it's all sorted now.

> Decommission ocp01 and ocp02 on Bytemark
I will action our standard decommissioning process. 
Shutting the instances down for two weeks and then terminating.

Would you like a final backup of anything on these servers? I assume everything we'd want is already in Github etc.

jpmckinney:
> This is to update anything that points to the cove review applications.

Aha, I've done that for our GitHub repositories - the updates to the standard documentation are in the next release (TBD). Otherwise, there are Google Docs, etc. that refer to the old URL, but we can change them as we find them.

> Would you like a final backup of anything on these servers?

I think everything is on GitHub or transferred to the new server :) (/media, db.sqlite3)

PR #413, opened 2023-03-07
Use official rabbitmq apt repos

### LINKED ISSUE #412: Upgrade RabbitMQ
Copying from Slack:

> The data registry (ocp13) is running RabbitMQ 3.8.2 and Erlang 22. RabbitMQ is now available at 3.11 and Erlang at 25, with many improvements to performance and reliability. We currently have regular application errors about weekly, and the new versions might fix the problem. Could we prioritize their upgrade? Here’s the [official upgrade docs](https://www.rabbitmq.com/upgrade.html).
> 
> In terms of options, you can forward the remote port 15672 to the local port 15673 (`ssh -N ssh://root@ocp13.open-contracting.org:2223 -L 15673:localhost:15672`) and then visit http://localhost:15673/#/queues (user/pass in deploy repo). If all the queues are empty with no activity, you can start (you should stop all kingfisher-process containers, all pelican-backend containers, and the exporter, flattener, wiper containers in the data-registry compose file). You can even just completely remove RabbitMQ and reinstall it (starting the containers will cause the exchanges, queues and bindings to be created). This option might be simplest, as otherwise it looks like you’d have to upgrade each minor version one-at-a-time https://www.rabbitmq.com/upgrade.html#rabbitmq-version-upgradability

with reply:

> The simplest approach indeed looks to be removing RabbitMQ and re-installing at the latest version - as you've mentioned above.
> I'll work on a PR moving to the cloudsmith apt repositories and schedule in the deployment for a quiet time.
https://www.rabbitmq.com/install-debian.html#apt-cloudsmith

RobHooper:
Fixes #412

### DIFF
--- a/salt/rabbitmq/init.sls
+++ b/salt/rabbitmq/init.sls
@@ -1,8 +1,31 @@
 # https://github.com/salt-formulas/salt-formula-rabbitmq/blob/master/rabbitmq/server/service.sls
 
+rabbitmq-dependancies:
+  pkg.installed:
+    - pkgs:
+      - gnupg2
+      - apt-transport-https
+
+erlang repo:
+  pkgrepo.managed:
+    - humanname: Erlang Official Repository
+    - name: deb http://ppa.launchpad.net/rabbitmq/rabbitmq-erlang/ubuntu {{ grains.oscodename }} main
+    - dist: {{ grains.oscodename }}
+    - file: /etc/apt/sources.list.d/rabbitmq_erlang.list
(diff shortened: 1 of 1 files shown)


RobHooper:
Go live process:
1. Merge into Master
2. Stop Registry application / docker containers.
3. Manually uninstall rabbitmq (`apt-get purge rabbitmq-server`)
4. Deploy salt for ocp13
5. Start the application (if not completed by step 4)

jpmckinney (approved):
Only change to go-live is to check that Rabbit queues are empty.

jpmckinney:
If it's just data_registry_production_flattener_init we can go ahead - those messages are easy to requeue.

Issue #402, opened 2023-01-11
Rebuild Kingfisher server


jpmckinney:
Moving from CRM-8007

It will be easier to deploy the new Kingfisher Process version with a new, empty database where all the new tables schemas can be applied instead of planning a migration for the current tables.

* [x] We can re-use most of the data registry's configuration.
* [x] Check whether non-OCP postgres users are still using their logins (point them to the data registry instead)
* [x] Remove the "Kingfisher personal SQL user account" issue template (and any links to it)

----

To do after PR merge:

* [x] Add Redash data source for new server
* [x] Merge upgrade-jsonschema in kingfisher-collect and change to main branch in this repo
* [x] Update Redash data source to use postgres.kingfisher.open-contracting.org (instead of ocp23)
* [x] Update collect.kingfisher CNAME
* [x] Delete process.kingfisher CNAME
* [x] Deploy docs server to apply tinyproxy changes to allowed IPs
* [x] I think we can not have the CNAME process.kingfisher and process1.kingfisher (we have enough with collect.kingfisher, postgres.kingfisher, etc.)
* [x] Notify korsikov/Yaroslav/Sergey and Vadim on Slack of the new domain name and database name (kingfisher_collect) for Qlik BI

jpmckinney:
See this milestone for some issues to ideally close before the new deployment: https://github.com/open-contracting/kingfisher-process/milestone/7

jpmckinney:
Related:

#142

jpmckinney:
Once this is done, I can also upgrade the requirements files for:

* [ ] kingfisher-collect (notably Scrapy and requests) https://github.com/open-contracting/kingfisher-collect/pull/1003 https://github.com/open-contracting/kingfisher-collect/issues/998
* [ ] kingfisher-summarize (notably requests) https://github.com/open-contracting/kingfisher-summarize/pull/217 (checking with team on https://github.com/open-contracting/kingfisher-summarize/issues/174)

jpmckinney:
Okay, Kingfisher Collect and Summarize are ready for Python 3.10.

To test the deployment, you can use the branches `upgrade-jsonschema` (Collect) and `190-filter-sql-json-path` (Summarize).

jpmckinney:
Instructions to share with data support managers:

* Kingfisher Process changes (local load)
* RabbitMQ management interface #430 (I think there is just an admin user, currently, not a read-only user)

PR #418, opened 2023-03-25
Kingfisher Process Docker app deploy

### LINKED ISSUE #200: Set up Pelican server
~~Once I've migrated Kingfisher Archive files to S3, I'll start setting up this server.~~ Just putting some notes down.

### LINKED ISSUE #427: Update create server process to not add port knocking if Docker to be added later
Right now we run onboarding and core first, which runs the firewall script. The firewall script checks for `command -v docker`, which will fail. It will then proceed to implement port knocking. https://ocdsdeploy.readthedocs.io/en/latest/deploy/create_server.html#apply-core-changes

We can either:
* Check for either `command -v docker` or a non-empty `DOCKER` variable (e.g. add `DOCKER` to the `firewall-settings.local` template). That way, we can add a step to the documentation to set `DOCKER`, if appropriate.
* Document how to reverse the firewall script. I did this by flushing iptables, adding port 22, deploying, and restarting the server. If that's it, we can document the steps more explicitly.

jpmckinney:
closes #200, addresses #402, closes #427

Main:

* [x] Remove configuration for Kingfisher Process v1
* [x] @yolile will delete some of the DB users
  * If any others are removed from Pillar data after June 9, remember to run states to delete the SQL users
* [x] Copy things over for RBC Qlik BI: the `ocdskingfishercollect` DB (to be renamed `kingfisher_collect`) and `/home/collect/data` directory (to be moved to `/home/incremental/data`). We would need to comment-out the cron jobs until those two have been moved.
* [x] Once the `kingfisher_collect` database and `/home/incremental/data` directory are populated, change `cron.absent` to `cron.present` in `incremental.sls`
* [x] @RobHooper to add network configuration, review bullets below, etc.
* [x] Decide on `set -euo pipefail` for `firewall_reset.sh` https://github.com/open-contracting/deploy/pull/418#discussion_r1221830908

Notes:

* I haven't checked pgbackrest at all, e.g. `postgres.backup` in Pillar, `salt/postgres/files/pgbackrest`, etc. Do we need to change the pgbackrest configuration (kingfishser_common.sls) to avoid conflict with the old DB backups?
* I think we'll want to remove ocp04 from Salt, but not sure what we'll do about ocp05 (current replica), since we want to keep it up for up to a year at most, in case we want to access that database.
* I haven't checked replication at all.

----

* [x] Add these steps to `create_server.rst` under "5. Migrate from the old server" once clarified whether needed

Once main server is online:

* [x] Check uid of docker user patches uid in pillar/kingfisher_process.sls
* [x] Update tinyproxy IPs in pillar/tinyproxy.sls
* [x] Re-enable backups (restore configuration commented out in 9fca6f362115ff830f8b281eb9542ba8a2a1d49a and https://github.com/open-contracting/deploy/pull/418/commits/13321860c6c0ad1a8ac6bab6ba0fbb394188ae79 and in pillar/private/kingfisher_common.sls)
* ~~Update firewall IPs in pillar/kingfisher_replica.sls~~

Once replica server is online:

* ~~Update replica_ipv4 (and replica_ipv6) in pillar/kingfisher_main.sls~~

One either server is online:

* [x] Run https://github.com/jfcoz/postgresqltuner to see if any changes necessary
* [x] Double-check vm.nr_hugepages https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-HUGE-PAGES
  * This is based on peak memory usage (VmPeak from /proc/123/status). That said, the new server has 64GB not 128GB RAM, and postgresqltuner recommends halving this number, so I did that.


jpmckinney:
@RobHooper I've made the changes I wanted to make. There is a corresponding PR for private Pillar.

jpmckinney:
If you upgrade to Salt 3006 for this work, please check whether `{% include 'postgres/files/conf/shared.include' %}` still works (I opened #432 for a possibly related issue).

jpmckinney:
For Ubuntu Pro, the only server option seems to be "Physical servers with unlimited VMs" (we have no VMs).

Would we need to go with that one, or is "Desktop" the option to take?

I'm happy with either kingfisher-2023 or kingfisherv15. We can go with kingfisher-2023 :)

RobHooper:
The "Ubuntu Pro Server with unlimited VMs" option is indeed the best option for us because it is a physical server.
We need Ubuntu Universe repository coverage included because we are using a number of packages installed from there.

It should cost $500 for the year.

salt/prometheus/files/conf-prometheus.yml
```diff
@@ -16,7 +16,7 @@ scrape_configs:
     'cove-ocds-2': 'ocp18.open-contracting.org',
     'credere': 'ocp22.open-contracting.org',
     'ocds-kingfisher-replica': 'ocp05.open-contracting.org',
-    'ocds-kingfisher2': 'ocp04.open-contracting.org',
+    'ocds-kingfisher': 'ocp23.open-contracting.org',
```
RobHooper, line 19: @jpmckinney I'd like to standardise Prometheus so that it matches the salt roster (`kingfisher-main`), unless there is a reason not to?

RobHooper:
**Network Configuration**
I have not added a network configuration because it doesn't need changing.
We had been configuring networking deliberately on Linode servers to specifically configure IPv6 support, this is not needed on this server because the out of the box Hetzner configuration is fine.

**Tiny Proxy**
I cannot see how Tiny Proxy is in use.
The configuration it was first committed with, such as `KINGFISHER_PROXY_URL`, no longer exists. 
Is Tiny Proxy still needed?

salt/prometheus/files/conf-prometheus.yml
jpmckinney: Fine with me :)

salt/core/firewall/files/firewall_reset.sh
```diff
@@ -0,0 +1,34 @@
...
+iptables -t mangle -X
+iptables -t raw -F
+iptables -t raw -X
+ip6tables -P INPUT ACCEPT
+ip6tables -P FORWARD ACCEPT
+ip6tables -P OUTPUT ACCEPT
+ip6tables -F
+ip6tables -X
+ip6tables -Z
+ip6tables -t nat -F
```
jpmckinney, line 29: I noticed that `firewall.sh` doesn't reset nat for ipv6. Should that be added?

salt/core/firewall/files/firewall_reset.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# This file is managed by Dogsbody Technology Ltd.
+#   https://www.dogsbody.com/
+#
+# Description:  A script to reset iptables rules.
+#
+# Usage:  $0
+#
+
+iptables -P INPUT ACCEPT
```
jpmckinney, line 11: Should this file `set -euo pipefail`?

Also, should it `source /home/sysadmin-tools/firewall-settings.local` to use variables for the command paths like in `firewall.sh`, or is that generally unnecessary?

jpmckinney:
> **Tiny Proxy**
> I cannot see how Tiny Proxy is in use.
> The configuration it was first committed with, such as `KINGFISHER_PROXY_URL`, no longer exists.
> Is Tiny Proxy still needed?

Tinyproxy is running on `docs`. Scrapyd gets the tinyproxy URL in `scrapyd.service`

```
Environment="http_proxy={{ pillar.tinyproxy.url }}"
Environment="https_proxy={{ pillar.tinyproxy.url }}"
```

salt/core/firewall/files/firewall_reset.sh
RobHooper: Added in new commit.

salt/kingfisher/summarize/init.sls
```diff
@@ -1,22 +1,48 @@
-{% from 'lib.sls' import set_cron_env %}
+{% from 'lib.sls' import create_user, set_cron_env %}
 
 include:
-  - kingfisher.process
+  - python.psycopg2
   - python_apps
```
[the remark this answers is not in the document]
jpmckinney: For Summarize, python_apps is installing python virtualenv, cloning git, creating a virtualenv, installing requirements, and adding a config file.

I think it's better to just make the uwsgi and apache `include`s conditional in python_apps.

salt/kingfisher/collect/incremental.sls
```diff
@@ -1,5 +1,4 @@
 {% from 'lib.sls' import create_user, set_cron_env %}
-{% from 'kingfisher/collect/init.sls' import directory as scrapyd_directory %}
 
 include:
```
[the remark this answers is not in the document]
jpmckinney: For incremental.sls, python_apps is installing python virtualenv, cloning git, creating a virtualenv, and installing requirements.

Similar to previous comment, we should preferably just make the includes in `python_apps` conditional.

jpmckinney:
@RobHooper Ah, the inclusion of those modules is conditional on data from the for loop. I'll try something out.

jpmckinney:
@RobHooper Okay, I committed a solution.

RobHooper:
@jpmckinney , I've deployed this PR to new ocp23 server and fixed a few bugs that flagged up.
I disabled replication for the deployment so that we are not blocked by the replica server build (but I have not committed this since it is not permanent).

There are three errors outstanding from the deployment:
1
> ID: grant kingfisher_summarize_read schema privileges in summaries
> Function: postgres_privileges.present
> Name: kingfisher_summarize_read
> Result: False
> Comment: Failed to grant privilege(s): USAGE to kingfisher_summarize_read

2
> ID: grant kingfisher_summarize_read table privileges in summaries
> Function: postgres_privileges.present
> Name: kingfisher_summarize_read
> Result: False
> Comment: Failed to grant privilege(s): SELECT to kingfisher_summarize_read

3
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> pip-tools 6.13.0 requires click>=8, but you have click 7.1.2 which is incompatible.

There is also a number of states that are re-running on subsequent runs that checking and addressing.

jpmckinney:
We're now using V2 on credere and kingfisher-main.

We're still installing it as [standalone](https://docs.docker.com/compose/install/) at the bottom of docker/init.sls.

The Docker docs have:

> This install scenario is not recommended and is only supported for backward compatibility purposes.

@RobHooper Should we remove that state? `docker compose` (without hyphen) is already available.

jpmckinney:
> but I have not committed this since it is not permanent

What changes do I need to make in order to do the deploy and see the errors myself?

jpmckinney:
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour > is the source of the following dependency conflicts.
> pip-tools 6.13.0 requires click>=8, but you have click 7.1.2 which is incompatible.

I merged the branch of kingfisher-summarize mentioned here: https://github.com/open-contracting/deploy/issues/402#issuecomment-1563769793

jpmckinney:
The repeated states I'm seeing are:

```
          ID: deployer_user_exists
    Function: user.present
        Name: deployer
      Result: True
     Comment: Updated user deployer
     Started: 16:41:03.483180
    Duration: 29.511 ms
     Changes:   
              ----------
              groups:
                  - deployer
```

```
          ID: add deployer user to docker group
    Function: group.present
        Name: docker
      Result: True
     Comment: The following group attributes are set to be changed:
              addusers: ['deployer']
     Started: 16:27:11.771557
    Duration: 4.064 ms
     Changes:   
              ----------
              Final:
                  All changes applied successfully
```

The states like `ID: grant kingfisher_collect_read table privileges in public` were being repeated on the old server as well. I thought I had seen an old issue related to that, but can't find it now.

jpmckinney:
I made a commit to fix the failing states.

salt/core/firewall/files/firewall_reset.sh
jpmckinney: I think the most likely scenario is someone makes a commit that introduces an error. Without `set -euo pipefail`, the command that errors will be silent, and it'll appear like the script succeeded. I think it's better to have feedback.

Since this script is always manually run while logged into the server, there's not much consequence to terminating early, as far as I can see.

jpmckinney:
I added ports 80,443 and 7231 to the Hetzner firewall for ocp23, reset the server-side firewall, and restarted Docker.

RobHooper:
> Should we remove [the docker compose standalone] state?

Agreed, we should move to docker compose v2, it gets installed via apt by default now so salt no longer needs manual configuration around this. This is probably it's own pull request so that nothing is missed.

Regarding Kingfisher, can we update the Pelican config to v2 and migrate the cron? (`pelican/backend/init.sls`).

salt/postgres/init.sls
```diff
@@ -195,6 +203,9 @@ grant {{ group }} schema privileges in {{ schema }}:
     - maintenance_db: {{ database }}
     - require:
       - postgres_database: {{ database }}
+  {% if schema != 'public' %}
+      - postgres_schema: {{ schema }}_sql_schema
```
RobHooper, line 207: > The following requisites were not found:
> require:
> postgres_schema: summaries_sql_schema

Also erroring for `reference_sql_schema`

RobHooper:
I've committed a change disabling Postgres backups so that you can test the deployment.
I am seeing errors with the new schema requirement added to the `grant _ table privileges in _` state. - See also comment against the code.

The `postgres_privileges.present` issue where it always presents as changed occurred on the previous server too, this will indeed be the bug with Salt you flagged.

jpmckinney:
Strange - I deployed the states on the weekend. I’ll have a look

salt/postgres/init.sls
RobHooper: Needed to update pillar/private

jpmckinney:
Actually – I'm not 100% sure we need a replica right away. Our usage patterns have changed somewhat since we originally created the replica. I don't think we've had such high load on the main server that it couldn't fulfill the replica load.

What do you think?

RobHooper:
Happy being led by you if we bring the replica online now or later.

We will loose the failover / disaster recovery that the replica server currently provides.
We have database backups but it is a slower process to recover with them.
The replica server is also significantly quicker to failover to in the case of hardware failures (i.e. the worst-case scenario).

I am also conscious that the new server is smaller so this is another factor to consider around load.

If we are not starting the replica now I will configure pgbackrest before this is merged.

jpmckinney:
Let's start without the replica.

salt/core/firewall/files/firewall_reset.sh
RobHooper: Added

pillar/kingfisher_main.sls
```diff
@@ -0,0 +1,201 @@
...
+    # "unable to remove file '%s'"
+    # https://github.com/pgbackrest/pgbackrest/blob/4adf6eed09da3f0819abef813c5a44deb9c91487/src/storage/storage.intern.h#L43
+    # "expire command encountered %u error(s), check the log file for details"
+    # https://github.com/pgbackrest/pgbackrest/blob/4adf6eed09da3f0819abef813c5a44deb9c91487/src/command/expire/expire.c#L1078
+    cron: |
+        MAILTO=root
+        # Daily incremental backup
+        15 05 * * 0-2,4-6 postgres pgbackrest backup --stanza=kingfisher-2023
+        # Weekly full backup
+        15 05 * * 3 postgres pgbackrest backup --stanza=kingfisher-2023 --type=full 2>&1 | grep -v "unable to remove file.*We encountered an internal error\. Please try again\.\|expire command encountered 1 error.s., check the log file for details"
```
jpmckinney, line 127: I added some comments above to check whether this grep still matches text.

I think the `1` needs to be changed to `\d+` (in which case `grep` needs a `-E` option)

I couldn't find what outputs `We encountered an internal error\. Please try again\.`

Is the `\|` meant to match a literal pipe character? (I'm not sure how pgbackrest concatenates messages).

jpmckinney:
@RobHooper This PR is good to go. I added one comment, but it can be looked into separately.

PR #433, opened 2023-05-30
Salt 3006 Onedir support

RobHooper:
+ Added relative `Saltfile` configuration to prevent clashing with other salt deployments installed.
+ Following salt-ssh symlinks to find the python3 binary, this was broken on Debian based systems.


jpmckinney (approved):
I need Saltfile at `~/.salt/`, but I just updated instructions that this could be needed on macOS.

Issue #218, opened 2020-12-11
Apache default site 


RobHooper:
The Apache default site is what gets served when a request doesn't match a virtual host. 
This happens when visitors go to places that you don't expect, for example: the server IP address, a DNS record that isn't used to host a site (ocp02.open-contracting.org), a site we've switched off or if someone points a domain we have no control over at the server. 

The biggest risk with an absent default site is that requests to these endpoints returns one of our sites, this would lead to duplicate content which is bad for SEO. 

Currently HTTP traffic receives the Apache default site which is OK, it addresses the risks of this issue but it does look ugly. 

HTTPS however doesn't having a specific default site and needs addressing. 

Best practice is to serve a custom holding page for these requests or to redirect (302 redirect) to an actual site.

RobHooper:
Ideally, the holding page would also be served for all paths under the domain.
Especially if it is replacing a site that we've turned off.

Currently the apache default page is being returned for just the main FQDN, paths under it receive 404s.

jpmckinney:
When a disallowed host reaches Django, it responds with 400 Bad Request: https://docs.djangoproject.com/en/3.1/ref/exceptions/#django.core.exceptions.SuspiciousOperation

Can you link to some best practice? If best practice is to serve/redirect content, I'm fine with just redirecting everything to https://www.open-contracting.org

jpmckinney:
For HTTPS, I assume we don't need certs – but for any traffic that disregards the lack of certs, we can still respond.

RobHooper:
I don't have any links regarding best practices for how to configure the default site I'm afraid. 
It is common practice to disable the inbuilt Apache page.
It is bad to serve the same site on multiple domains as this is duplicate content.

For HTTPS we'll need a cert but we're not worried about it being valid because no legitimate traffic should end up here. We often use an SSL cert matching the server hostname or a self signed one.

jpmckinney:
I was curious whether it's better to error with HTTP 400 (simplest), serve some holding page or redirect to another site.

RobHooper:
I'm happy with returning an error code, probably 404 or 410 depending on the link. 
In NGINX it has a "444" status that just drops connection and returns nothing to the client, I've set that up as the default site before. Not sure what the Apache equivalent is.

If we switch off a site and it falls back to this it might be nice to send visitors a "OCP" page rather than anything systematic.

jpmckinney:
I think 444 is nginx-specific. I'm happy with 404.

[...]

PR #434, opened 2023-05-31
Apache defaults

dogsbody-ashley:
This PR updates the default Apache page to a custom page fixing issue [#218](https://github.com/open-contracting/deploy/issues/218)

The custom page HTML can be edited as required. It currently looks like this:

image


salt/apache/init.sls
```diff
@@ -33,6 +33,23 @@ apache2:
     - require:
       - pkg: apache2
 
+# Uploads custom Apache defaults page and configuration
+  file.managed:
+    - names: 
+      - /var/www/html/error_page.html:
+        - source: salt://apache/files/docs/error_page.html
+      - /etc/apache2/sites-avaliable/default.conf:
```
jpmckinney, line 41:
```suggestion
      - /etc/apache2/sites-available/default.conf:
```

salt/apache/init.sls
```diff
@@ -33,6 +33,23 @@ apache2:
...
+        - source: salt://apache/files/conf/default.conf
+    - require:
+      - pkg: apache2
+    - watch_in:
+      - service: apache2
+# Removes Apache defaults page and configuration
+/var/www/html/index.html:
+   file.absent
+/etc/apache2/sites-enabled/000-default.conf:
```
jpmckinney, line 50: Add newlines between states for readability.

```suggestion

# Remove Apache default page.
/var/www/html/index.html:
   file.absent

# Remove Apache default configuration.
/etc/apache2/sites-enabled/000-default.conf:
```

salt/apache/init.sls
```diff
@@ -33,6 +33,23 @@ apache2:
     - require:
       - pkg: apache2
 
+# Uploads custom Apache defaults page and configuration
+  file.managed:
+    - names: 
+      - /var/www/html/error_page.html:
+        - source: salt://apache/files/docs/error_page.html
```
jpmckinney, line 40: Suggest moving this from `files/docs` to just `files`, since `docs` is specific to ocp19.

salt/apache/init.sls
```diff
@@ -33,6 +33,23 @@ apache2:
...
+      - /etc/apache2/sites-avaliable/default.conf:
+        - source: salt://apache/files/conf/default.conf
```
jpmckinney, line 42: Since default.conf is installed in sites-available, we should put this in `files/sites` instead of `files/conf`.

salt/apache/files/conf/default.conf
```diff
@@ -0,0 +1,32 @@
...
+	# redirection URLs. In the context of virtual hosts, the ServerName
+	# specifies what hostname must appear in the request's Host: header to
+	# match this virtual host. For the default virtual host (this file) this
+	# value is not decisive as it is used as a last resort host regardless.
+	# However, you must set it for any further virtual host explicitly.
+	#ServerName www.example.com
+
+	ServerAdmin webmaster@localhost
+	DocumentRoot /var/www/html
+	ErrorDocument 404 error_page.html
```
jpmckinney, line 13: Since this document is specific to 404, let's rename it 404.html

```suggestion
	ErrorDocument 404 error_page.html
```

salt/apache/files/conf/default.conf
```diff
@@ -0,0 +1,32 @@
...
+	# The ServerName directive sets the request scheme, hostname and port that
+	# the server uses to identify itself. This is used when creating
+	# redirection URLs. In the context of virtual hosts, the ServerName
+	# specifies what hostname must appear in the request's Host: header to
+	# match this virtual host. For the default virtual host (this file) this
+	# value is not decisive as it is used as a last resort host regardless.
+	# However, you must set it for any further virtual host explicitly.
+	#ServerName www.example.com
+
+	ServerAdmin webmaster@localhost
```
jpmckinney, line 11:
```suggestion
	ServerAdmin {{ pillar.system_contacts.root }}
```

salt/apache/files/docs/error_page.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
jpmckinney, line 1: We'd probably do something more like this (removing the meta-refresh, inlining the minimum styles, and changing the message): https://github.com/open-contracting/extension-explorer/blob/main/extension_explorer/templates/redirect.html

jpmckinney (commented):
I reviewed the PR, but then I realized this is about changing the behavior of Apache when a user requests, for example: `ocp19.open-contracting.org`. It's not about changing the look of the default 404 page, which is something else.

The last comment in the issue links to a potential solution that just drops the request if the wrong Host is requested.

PR #441, opened 2023-06-28
postgres: Use Let's Encrypt certificates for PostgreSQL if postgres.ssl.servername is set

### LINKED ISSUE #440: Use Let's Encrypt certificates for PostgreSQL on ocp23
PowerBI's PostgreSQL connector doesn't like using the self-signed snakeoil certificate.

We have `postgres.kingfisher.open-contracting.org`. We can set up a dummy virtual host for it, and then otherwise follow: https://loganmarchione.com/2020/10/securing-postgres-connections-using-lets-encrypt-certificates/

jpmckinney:
closes #440 

Note that MDCertificateAuthority is set to https://acme-staging-v02.api.letsencrypt.org/directory in pillar/kingfisher_main.sls so that we can remove and get new certificates without going over the rate limit.

I think I have everything configured correctly, and running:

```bash
sudo -u www-data /opt/postgresql-certificates-wrapper.sh "postgres.kingfisher.open-contracting.org"
```

yields the correct result. However, this either isn't being run by MDNotifyCmd, or it's encountering an error (but I don't know how to find a relevant log).

### DIFF
--- a/pillar/kingfisher_main.sls
+++ b/pillar/kingfisher_main.sls
@@ -48,7 +48,6 @@ apache:
   modules:
     mod_md:
       MDNotifyCmd: /opt/postgresql-certificates-wrapper.sh
-      MDCertificateAuthority: https://acme-staging-v02.api.letsencrypt.org/directory
   sites:
     kingfisher-collect:
       configuration: proxy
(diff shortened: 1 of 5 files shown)


jpmckinney:
Ah, I figured it out: when getting the certs for the first time, they go in `staging`, not `domains`.

RobHooper:
I have configured SSL monitoring on postgres.kingfisher.open-contracting.org and PostgreSQL directly, this will alert us in-case the renewal failed silently. :+1:

PR #445, opened 2023-07-03
kingfisher: mount users directory to ease local loads

yolile:
I used the same directory structure in the containers so that we can use the current command and approach without changes

### DIFF
--- a/salt/docker_apps/files/kingfisher_process.yaml
+++ b/salt/docker_apps/files/kingfisher_process.yaml
@@ -11,10 +11,10 @@ x-filesystem: &filesystem
   volumes:
     # The right-hand path must match the settings.KINGFISHER_COLLECT_FILES_STORE default value in kingfisher-process.
     - {{ pillar.kingfisher_collect.env.FILES_STORE }}:/data
-    # For processing local loads.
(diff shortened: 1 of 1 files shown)


salt/docker_apps/files/kingfisher_process.yaml
```diff
@@ -11,6 +11,10 @@ x-filesystem: &filesystem
   volumes:
     # The right-hand path must match the settings.KINGFISHER_COLLECT_FILES_STORE default value in kingfisher-process.
     - {{ pillar.kingfisher_collect.env.FILES_STORE }}:/data
+    # For processing local loads.
+    {% for user, authorized_keys in pillar.users.items() %}
+    - /home/{{user}}/local-load:/home/{{user}}/local-load
+    {% endfor %}
```
jpmckinney, line 17: Change to a Jinja comment, otherwise this comment will appear on the registry server, without anything beneath it.

Change pillar.users to optional, since the registry server doesn't have this.

```suggestion
    {# To run the load command. #}
    {%- for user in pillar.users|default({}) %}
    - /home/{{user}}/local-load:/home/{{user}}/local-load
    {%- endfor %}
```

PR #452, opened 2023-08-07
Add IPv6 addresses to Linode servers

dogsbody-rob:
@jpmckinney Simple PR adding the new IPv6 blocks we have provisioned.

I am conscious that _if_ there are any errors / issues with this change we will cause an outage.

When is the best time to update and restart networking on these instances?
I was planning on scheduling in this deployment for ~09:00 BST one morning before America wakes up.


jpmckinney (approved):

jpmckinney:
You can do them all anytime except Credere. For that, @yolile will need to confirm.

PR #459, opened 2023-08-23
feat: disable apache version numbers

dogsbody-rob:
@jpmckinney, do we have any standardisation around config file names?

This config file needs to be applied after the other files because it overwrites some of their settings – thus zz-customization.

### DIFF
--- a/salt/apache/init.sls
+++ b/salt/apache/init.sls
@@ -85,16 +85,21 @@ disable site 000-default.conf:
   file.absent:
     - name: /var/www/html/index.html
 
+# Ensure this configuration is loaded after security.conf, provided by the package.
 /etc/apache2/conf-available/zz-customization.conf:
   file.managed:
     - contents: |
        ServerTokens Prod
        ServerSignature Off
+    - require:
+      - pkg: apache2
+    - watch_in:
+      - module: apache2-reload
 
 enable customization config:
   apache_conf.enabled:
     - name: zz-customization
-    - watch:
+    - require:
       - file: /etc/apache2/conf-available/zz-customization.conf
     - watch_in:
       - module: apache2-reload
--- a/salt/apache/init.sls
+++ b/salt/apache/init.sls
@@ -96,7 +96,7 @@ disable site 000-default.conf:
     - watch_in:
       - module: apache2-reload
 
-enable customization config:
+enable conf zz-customization.conf:
   apache_conf.enabled:
     - name: zz-customization
     - require:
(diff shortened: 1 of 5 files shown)


jpmckinney:
> This config file needs to be applied after the other files because it overwrites some of their settings

Which files are getting overwritten?

> do we have any standardisation around config file names?

Not explicitly, but we typically use numbers, like 99.

salt/apache/init.sls
```diff
@@ -85,6 +85,20 @@ disable site 000-default.conf:
   file.absent:
     - name: /var/www/html/index.html
 
+/etc/apache2/conf-available/zz-customization.conf:
+  file.managed:
+    - contents: |
+       ServerTokens Prod
+       ServerSignature Off
+
```
jpmckinney, line 93: I figure follow same requirement/dependency pattern as other `/etc/apache2/` files, unless the alternative in this PR is an improvement?

```suggestion
    - require:
      - pkg: apache2
    - watch_in:
      - module: apache2-reload

```

dogsbody-rob:
The file we're clashing with is `security.conf` which also modifies these settings. 
I prefer creating a new file over disabling or modifying this file because `security.conf` is managed by and will receive changes via apt.

Naming the new file 99- would put it first in the list because no other files in this directory are numbered.

Regarding letsencrypt, it's position in the order doesn't matter so it applies where-ever it falls alphabetically.

salt/redash/init.sls
```diff
@@ -0,0 +1,12 @@
+include:
+  - docker_apps
+
+/data/deploy/redash/files/nginx-security.conf:
```
jpmckinney, line 4:
```suggestion
{% from 'docker_apps/init.sls' import docker_apps_directory %}

include:
  - docker_apps

{% set entry = pillar.docker_apps.redash %}
{% set directory = docker_apps_directory + entry.target %}

{{ directory }}/files/nginx-security.conf:
```

salt/apache/init.sls
```diff
@@ -85,6 +85,20 @@ disable site 000-default.conf:
   file.absent:
     - name: /var/www/html/index.html
 
+/etc/apache2/conf-available/zz-customization.conf:
```
jpmckinney, line 88:
```suggestion
# Ensure this configuration is loaded after security.conf, provided by the package.
/etc/apache2/conf-available/zz-customization.conf:
```

salt/docker_apps/files/redash.yaml
```diff
@@ -18,6 +18,8 @@ services:
       - server
     links:
       - server:redash
+    volumes:
+      - /data/deploy/redash/files/nginx-security.conf:/etc/nginx/conf.d/security.conf
```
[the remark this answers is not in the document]
dogsbody-rob: Resolved in new commit.

jpmckinney (approved):

PR #460, opened 2023-08-24
Non ocds data

yolile:


salt/top.sls
```diff
@@ -42,6 +42,7 @@ base:
     - kingfisher
     - kingfisher.collect
     - kingfisher.collect.incremental
+    - kingfisher.non_ocds_collect
```
jpmckinney, line 45: Maybe we can call it `collect_generic`? (And in general use "generic" instead of "non-OCDS")

yolile: Done, but I have a problem. The GitHub URL is https://github.com/open-contracting/data-support.git, but the project and the requirements document are under the non_ocds_collect directory of that repo, I'm not sure how to indicate where the requirements document is in Salt

jpmckinney: It might be simpler/better to make a separate collect_generic repository.

PR #463, opened 2023-08-29
feat: add site for generic downloads

yolile:


jpmckinney (commented):
I think you can delete mod_autoindex.directory along with the related comment?

Otherwise, looks good (and hopefully works ;)).

PR #464, opened 2023-08-30
docs: standardise DNS TTLs

dogsbody-rob:


jpmckinney (approved):
I updated the TTLs for A and AAAA in create_server.rst to match the new TTLs in dns.rst

dogsbody-rob:
I have updated the ocpXX DNS TTLs in GoDaddy to match the new standard.

PR #465, opened 2023-08-31
Apache default improvements

dogsbody-rob:
The default site config needs to be the first site in-order to apply correctly.
If the default site is not first then another site will return to "unexpected traffic" such as direct requests to the IP address.
This bug exists on Cove because cove is first alphabetically.

- [ ] Delete default.conf in sites-enabled and file in sites-available

### DIFF
--- a/salt/apache/files/sites/_common.conf
+++ b/salt/apache/files/sites/_common.conf
@@ -34,5 +34,12 @@
     {%- if port == 443 or not ssl or not servername %}
     Include {{ includefile }}
     {%- endif %}
+
+    {%- if not servername and ssl and port == 80 %}
+    <IfModule mod_rewrite.c>
+      RewriteEngine On
+      RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
+    </IfModule>
+    {%- endif %}
 </VirtualHost>
 {%- endfor %}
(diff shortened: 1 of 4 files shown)


dogsbody-rob:
@jpmckinney, I would like to address the default site missing HTTP -> HTTPS redirect issue discussed over email, please can you advise on the best approach.

My current thought is a query such as `if name = "" and ssl and port = 80` then manually enter a redirect configuration (see below).
We don't want to enable mod_md because they shouldn't have a valid cert.
I am also concerned that `_common.conf` is very jinja heavy so this will make the file less readable.

```
RewriteEngine On
RewriteRule ^(.*)$ https://%{HTTP_HOST}$1 [R=301,L]
```

dogsbody-rob:
> > I would like to address the default site missing HTTP -> HTTPS redirect issue discussed over email, please can you advise on the best approach.
> 
> Can you remind me what the issue is?

SecurityScorecard is rating us F because the default pages are currently served over HTTP.
The default page has no value for an attacker but it would be nice to address to bring the score up.

jpmckinney:
I think before the closing VirtualHost:

```jinja
    {%- if not servername and ssl and port == 80 %}
    RewriteEngine On
    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
    {%- endif %}
```

jpmckinney (approved):

Issue #369, opened 2022-04-13
Docker Compose v1 to v2


dogsbody-ashley:
`Docker-compose` version 1 is End Of Life.
`Docker compose` version 2 should be installed and configured on all servers using it (currently ocp08, ocp13 & ocp14).
[More info in the Docker Compose docs.](https://docs.docker.com/compose/cli-command)

Updating to the new version **will** cause problems for **all** scripts running `docker-compose`, which has been replaced with `docker compose` (no hyphen). These scripts will need to be updated as well as the new install / update process via SALT. 

Docker offer a backwards compatible solution (`compose-switch`), but we've had issues implementing this on other servers so it may be easier to update directly. [Docker Compose Switch](https://docs.docker.com/compose/cli-command/#compose-switch).

jpmckinney:
Where is it described as end-of-life? The first link says "docker-compose is still supported and maintained" and "We currently don’t have a concrete timeline to deprecate Compose V1."

It also states "We are currently working towards providing a standard way to install Compose V2 on Linux. When this is available, Compose V2 will be marked as Generally Available (GA)."

With respect to replacing scripts, it also says "We are not planning to remove the aliasing of docker-compose to docker compose. We would like to make it easier for users to switch to V2 without breaking any existing scripts."

I'd prefer to wait for the standard way to install Compose V2.

The related issue in Docker is https://github.com/docker/roadmap/issues/257

dogsbody-ashley:
Hi James,

Happy to wait for Version 2 to be GA, 
Docker-Compose v1 has not had any updates since 10th May 2021 but we cannot see it having any security vulnerabilities and will continue to check for these.

jpmckinney:
The linked issue has been updated:

> On April 26, 2022, we announced the [GA of Docker Compose V2](https://www.docker.com/blog/announcing-compose-v2-general-availability/). We want you to have ample time to transition to Compose V2. We won’t sunset Docker Compose V1 immediately, and developers can still revert to V1. Given the numerous successful transitions to Compose V2 so far, we’ve created the following proposed timeline for Docker Compose V1's end of life (EOL):
>
> October 2022 - 6 Months Post GA
>
> * Support of critical bug fixes and severe security issues will end on Compose v1
> * Users can alias docker-compose to docker compose
> * Users can opt-out of V2 via the Docker Desktop UI or through the docker-compose disable-v2 command
>
> April 2023 - 1 Year Post GA
>
> * Users can alias docker-compose to docker compose
[...]

PR #466, opened 2023-09-01
New Credere production server

dogsbody-rob:
@yolile new Credere server is ready and running. 
Wanted to make you aware of some changes:
- The dev server has been renamed to `credere-staging` and the new server is now `credere`
- I have removed `docker-compose`, `docker compose` (space instead of hyphen) is installed and preferred. (This will be addressed everywhere under #369)
- Currently waiting for Linode to provide the new IPv6 range, should be configured in the next day or so.

Please can you review the new server names specifically, I wasn't sure what FQDN to use.


yolile (approved):

yolile (commented):
Maybe credere_dev instead of credere_staging? As we already have a credere_dev.sls file

dogsbody-rob:
@yolile Happy renaming bits from staging to dev.
Do the developers SSH keys in `credere_dev.sls` need access to both / the prod server?

yolile:
> Do the developers SSH keys in credere_dev.sls need access to both / the prod server?

No, only to dev

PR #492, opened 2024-04-22
Add PHP site error log

### LINKED ISSUE #454: Enable PHP-FPM error log

RobHooper:
closes #454.


salt/php-fpm/files/default.conf
```diff
@@ -6,3 +6,5 @@ listen.owner = {{ listen_user }}
 listen.group = {{ listen_user }}
 pm = static
 pm.max_children = 10
+php_admin_flag[log_errors] = on
```
RobHooper, line 9: I have coded error logs to be always on.
Is there a situation we may want to disable them?

salt/logrotate/files/php-site-logs
```diff
@@ -0,0 +1,13 @@
+/var/log/php-fpm/*/*.log {
+	rotate 14
+	daily
```
RobHooper, line 3: Storing daily logs for 14 days. This matches the Apache configuration.

salt/php-fpm/files/default.conf
jpmckinney: I can't think of one :)

salt/php-fpm/init.sls
```diff
@@ -22,7 +22,32 @@ php-fpm-reload:
...
+  file.managed:
+    - name: /etc/logrotate.d/php-site-logs
+    - source: salt://logrotate/files/php-site-logs
+    - template: jinja
+    - context:
+        php_version: {{ php_version }}
+    - require:
+      - pkg: php-fpm
+      - file: /var/log/php-fpm
+
```
jpmckinney, line 40: We have this existing process for logrotate files: https://ocdsdeploy.readthedocs.io/en/latest/develop/update/logs.html#logrotate

Maybe we just add the option for a context (php_version here) to be provided to that existing state?

I don't think we need to replicate the `require`, since if php-fpm, etc. aren't installed, then there won't be any php-fpm log files to match.

RobHooper: I've pushed a modification extending the logrotate state to support custom contexts.
If you're happy with this approach I will update kingfisher and the docs.

jpmckinney: PR looks good!

jpmckinney:
@RobHooper I see you made commits. Do you intend for me to re-review?

RobHooper:
> @RobHooper I see you made commits. Do you intend for me to re-review?

Happy for you to re-review, this PR is ready to deploy from my side.

Once this PR is approved, I will merge and deploy.

Issue #420, opened 2023-04-05
PostgreSQL backups (ocp13, ocp24)


ghost:
Linode provides file backups which is fine for site files however this is not ideal for Postgres.
During the backup Postgres will still have data loaded  into memory so we cannot ensure the quality of a backup using this approach.

We should use pgBackRest to store backups in AWS S3, as we have done on other servers
https://ocdsdeploy.readthedocs.io/en/latest/develop/update/postgres.html#set-up-backups

pgBackRest has the added benefit that it is offsite and recoverable without Linode.

----

Other non-development servers with postgres but without pgbackrest:

* [x] Credere (ocp24)
* [ ] Registry (ocp13) (see also #289)
  * Only the data_registry and spoonbill_web databases. (kingfisher_process, pelican_backend and pelican_frontend have transient data only.)

RobHooper:
Credere production backups setup:
https://github.com/open-contracting/deploy/commit/a86c0d50cabe8038a31bf50e62b3ae82cc616fe8

jpmckinney:
Just curious how the value `max_wal_senders = 4` is decided?

RobHooper:
@jpmckinney ,  `max_wal_senders` limits the number of concurrent connections streaming WAL data, usually this is to configure database replication but it is also used for our backups.
Theoretically we should only need `1` concurrent connection but most sources recommend `3`. `3` allows admins to add replication easily in the future if needed, it also leaves room if a connections hang.

I set `4` for Credere as an educated estimate being more limited than the default value (`10`) but happy to reduce.

PR #512, opened 2024-07-03
Add per-database PostgreSQL backups

RobHooper:
Addressing #420

### DIFF
--- a/salt/postgres/backup.sls
+++ b/salt/postgres/backup.sls
@@ -6,16 +6,6 @@ include:
 
 {{ set_config('aws-settings.local', 'S3_DATABASE_BACKUP_BUCKET', pillar.postgres.backup.location) }}
 
-set BACKUP_DATABASES setting:
-  file.keyvalue:
-    - name: /home/sysadmin-tools/aws-settings.local
-    - key: BACKUP_DATABASES
-    - value: '( "{{ pillar.postgres.backup.databases|join('" "') }}" )'
-    - append_if_not_found: True
-    - require:
-      - file: /home/sysadmin-tools/bin
-      - sls: aws
-
 /home/sysadmin-tools/bin/postgres-backup-to-s3.sh:
   file.managed:
     - source: salt://postgres/files/postgres-backup-to-s3.sh
@@ -30,6 +20,16 @@ set BACKUP_DATABASES setting:
         45 04 * * * root /home/sysadmin-tools/bin/postgres-backup-to-s3.sh
     - require:
       - file: /home/sysadmin-tools/bin/postgres-backup-to-s3.sh
+
+set BACKUP_DATABASES setting:
+  file.keyvalue:
+    - name: /home/sysadmin-tools/aws-settings.local
+    - key: BACKUP_DATABASES
+    - value: '( "{{ pillar.postgres.backup.databases|join('" "') }}" )'
+    - append_if_not_found: True
+    - require:
+      - file: /home/sysadmin-tools/bin
+      - sls: aws
 {% elif salt['pillar.get']('postgres:backup:type') == 'pgbackrest' %}
 include:
   - postgres
--- a/salt/postgres/backup.sls
+++ b/salt/postgres/backup.sls
@@ -1,6 +1,10 @@
+{% from 'lib.sls' import set_config %}
+
 include:
   - postgres
+  - aws
 
+{% if salt['pillar.get']('postgres:backup:type') == 'pgbackrest' %}
 # Require official PostgreSQL repo as they provide a newer version of pgbackrest.
 pgbackrest:
   pkg.installed:
@@ -10,16 +14,27 @@ pgbackrest:
 
 /etc/pgbackrest/pgbackrest.conf:
   file.managed:
-   - makedirs: True
-   - user: postgres
-   - group: postgres
-   - source: salt://postgres/files/pgbackrest/{{ pillar.postgres.backup.configuration }}.conf
-   - template: jinja
+    - makedirs: True
+    - user: postgres
+    - group: postgres
+    - source: salt://postgres/files/pgbackrest/{{ pillar.postgres.backup.configuration }}.conf
+    - template: jinja
+
+{% elif salt['pillar.get']('postgres:backup:type') == 'script' %}
+{{ set_config('aws-settings.local', 'S3_DATABASE_BACKUP_BUCKET', pillar.postgres.backup.location ) }}
+{{ set_config('aws-settings.local', 'BACKUP_DATABASES', ' '.join(pillar.postgres.backup.databases) ) }}
+
+/home/sysadmin-tools/bin/postgres-backup-to-s3.sh:
+  file.managed:
+    - source: salt://postgres/files/postgres-backup-to-s3.sh
+    - mode: 750
+    - require:
+      - file: /home/sysadmin-tools/bin
+      - sls: aws
+{% endif %}
 
 {% if salt['pillar.get']('postgres:backup:cron') %}
 /etc/cron.d/postgres_backups:
   file.managed:
     - contents_pillar: postgres:backup:cron
-    - require:
-      - pkg: pgbackrest
 {% endif %}
--- a/salt/postgres/backup.sls
+++ b/salt/postgres/backup.sls
@@ -5,7 +5,16 @@ include:
   - aws
 
 {{ set_config('aws-settings.local', 'S3_DATABASE_BACKUP_BUCKET', pillar.postgres.backup.location) }}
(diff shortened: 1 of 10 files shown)


salt/postgres/backup.sls
```diff
@@ -10,16 +14,27 @@ pgbackrest:
...
-   - source: salt://postgres/files/pgbackrest/{{ pillar.postgres.backup.configuration }}.conf
-   - template: jinja
+    - makedirs: True
+    - user: postgres
+    - group: postgres
+    - source: salt://postgres/files/pgbackrest/{{ pillar.postgres.backup.configuration }}.conf
+    - template: jinja
+
+{% elif salt['pillar.get']('postgres:backup:type') == 'script' %}
+{{ set_config('aws-settings.local', 'S3_DATABASE_BACKUP_BUCKET', pillar.postgres.backup.location ) }}
```
jpmckinney: This is the same variable used for MySQL backups. If we have a server with both MySQL and PostgreSQL, will the backups conflict, or will they be fine to share the bucket?

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,51 @@
...
+PSQL_ADMIN_USER="postgres"
+
+# shellcheck disable=SC1091
+. /home/sysadmin-tools/aws-settings.local
+
+export AWS_ACCESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION
+
+if [ "$LOGNAME" != "postgres" ]; then
```
jpmckinney, line 14: Isn't this script run by root's cron? (hence the `sudo -u postgres` calls)

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Create work directory resolving permissions alerts when running psql from root.
+WORKDIR=$(mktemp -d)
+chown $PSQL_ADMIN_USER:$PSQL_ADMIN_USER "$WORKDIR"
+
+mapfile -t DATABASES < <(cd "$WORKDIR" && sudo -u $PSQL_ADMIN_USER /usr/bin/psql -t --csv -c 'select datname from pg_database')
+
+# Using read over mapfile because the latter leaves a newline on the final item when processing a space delimited string.
+read -ra REQUESTED_DATABASES <<< "${BACKUP_DATABASES}"
+
```
jpmckinney: `salt/backup/init.sls` does something like this when writing the configuration, to avoid having to do this kind of `read` in `site-backup-to-s3.sh`

```yaml
set BACKUP_DIRECTORIES setting:
  file.keyvalue:
    - name: /home/sysadmin-tools/aws-settings.local
    - key: BACKUP_DIRECTORIES
    - value: '( "{{ pillar.backup.directories|join('" "') }}" )'
    - append_if_not_found: True
    - require:
      - file: /home/sysadmin-tools/bin
      - sls: aws
```

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ ! -x "$AWS_CLI" ]; then
+    echo "Error: The aws executable is not installed"
+    exit 3
+fi
+
+# Create work directory resolving permissions alerts when running psql from root.
+WORKDIR=$(mktemp -d)
+chown $PSQL_ADMIN_USER:$PSQL_ADMIN_USER "$WORKDIR"
+
+mapfile -t DATABASES < <(cd "$WORKDIR" && sudo -u $PSQL_ADMIN_USER /usr/bin/psql -t --csv -c 'select datname from pg_database')
```
jpmckinney: Instead of the `WORKDIR` workaround, why not `su - postgres -c "command"` ?

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Using read over mapfile because the latter leaves a newline on the final item when processing a space delimited string.
+read -ra REQUESTED_DATABASES <<< "${BACKUP_DATABASES}"
+
+for DATABASE in "${REQUESTED_DATABASES[@]}"; do
+    if [[ "${DATABASES[*]}" =~ $DATABASE ]]; then
```
jpmckinney, line 34: Instead of a condition, why not let `pg_dump` fail? I suppose the condition prevents `mktemp` from creating files that will not be cleaned up, but I think it's OK to create some unused temp files.

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+# Create work directory resolving permissions alerts when running psql from root.
+WORKDIR=$(mktemp -d)
+chown postgres:postgres "$WORKDIR"
+
+mapfile -t DATABASES < <(cd "$WORKDIR" && sudo -u postgres /usr/bin/psql -t --csv -c 'select datname from pg_database')
+
+# Using read over mapfile because the latter leaves a newline on the final item when processing a space delimited string.
+read -ra REQUESTED_DATABASES <<< "$BACKUP_DATABASES"
```
jpmckinney, line 31: `salt/backup/init.sls` has a different way of writing an array to aws-settings.local, such that `read` isn't needed:

```yaml
set BACKUP_DIRECTORIES setting:
  file.keyvalue:
    - name: /home/sysadmin-tools/aws-settings.local
    - key: BACKUP_DIRECTORIES
    - value: '( "{{ pillar.backup.directories|join('" "') }}" )'
    - append_if_not_found: True
    - require:
      - file: /home/sysadmin-tools/bin
      - sls: aws
```

salt/postgres/files/postgres-backup-to-s3.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ ! -x "$AWS_CLI" ]; then
+    echo "Error: The aws executable is not installed"
+    exit 3
+fi
+
+# Create work directory resolving permissions alerts when running psql from root.
+WORKDIR=$(mktemp -d)
+chown postgres:postgres "$WORKDIR"
+
+mapfile -t DATABASES < <(cd "$WORKDIR" && sudo -u postgres /usr/bin/psql -t --csv -c 'select datname from pg_database')
```
jpmckinney, line 28: Instead of the `WORKDIR` workaround, why not `su - postgres -c "command"` ?

RobHooper: Swapping user with `su -` clears the working environment which was creating problems in a previous version of this script (bash functions are not accessible after changing user). I ended up not using functions so this is not needed any more.

Updated as suggested.

salt/postgres/files/postgres-backup-to-s3.sh
RobHooper: Because we have `set -euo pipefail` if a backup fails it would stop the script from continuing preventing other database backups. Handling this situation lets us send a clear error message and the backups can continue.

I have not tested how `pg_dump` acts when given a malformed database name, I expect it returns a non-zero exit code. I hope `pg_dump` also returned a clear error message - which we would need if we were to remove this check.

salt/postgres/backup.sls
RobHooper: If we had both backup scripts on one server the backups would be muddled together and salt would clash setting this variable twice. 
Backup file names would be especially confusing if the databases had similar names between services, in this case, the only difference would be the file name ending either `.sql.gz` (MySQL) or `.tar.gz` (PostgreSQL).

However, it is very unlikely and it would be bad design to have MySQL and PostgreSQL on the same server since they both fill the same task in the stack - so this shouldn't be a problem.

salt/postgres/files/postgres-backup-to-s3.sh
RobHooper: Fixed in new commit.

salt/postgres/files/postgres-backup-to-s3.sh
RobHooper: We can indeed create the array in Salt for bash to use. I had overlooked this.
Fixed in new commit.

salt/postgres/files/postgres-backup-to-s3.sh
RobHooper: Fixed in new commit.

salt/postgres/files/postgres-backup-to-s3.sh
jpmckinney: Aha, good call.

FYI, `pg_dump foo` returns status code 1 and:

```
pg_dump: error: connection to database "foo" failed: FATAL:  database "foo" does not exist
```

jpmckinney (approved):

PR #519, opened 2024-08-21
credere: use new format for running commands

yolile:
to be merged with https://github.com/open-contracting/credere-backend/pull/380 and https://github.com/open-contracting/credere-backend/pull/381


jpmckinney:
I removed my accidental `-v`.

PR #523, opened 2024-09-17
dream: Update configuration

jpmckinney:


jpmckinney:
Synced to prod, not deployed (build versions don't match – can comment those out if we want to deploy the other changes)

PR #538, opened 2025-01-10
docs: note salt Python 3.12 bug

RobHooper:
@jpmckinney , just documenting a manual step I ran on `ocp27`.
Salt doesn't yet support Python 3.12 and needs the below workaround for deployments to work.

This bug only occurs for me on `ocp27` and only after upgrading my PC (both systems are now Ubuntu 24.04 and therefore Python 3.12).

Let me know if this is best documented somewhere else.

### DIFF
--- a/docs/deploy/create_server.rst
+++ b/docs/deploy/create_server.rst
@@ -513,6 +513,16 @@ Configure reverse DNS
 
       The server's Pillar file needs ``system_contacts``, ``network.domain``, ``ssh.admin``, ``locale``, ``ntp`` and, preferably, ``maintenance`` sections.
 
+   .. warning::
+
+      Salt deployments between Python 3.12 systems are `currently broken <https://github.com/saltstack/salt/issues/65360>`__, returning the error message: `ModuleNotFoundError: No module named 'backports'`.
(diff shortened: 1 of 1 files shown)


jpmckinney:
What version of Salt are you using? How did you install it? The issue seems to suggest that the onedir versions shouldn’t have the problem https://docs.saltproject.io/salt/install-guide/en/latest/topics/upgrade-to-onedir.html#upgrade-to-onedir

RobHooper:
> What version of Salt are you using? How did you install it? The issue seems to suggest that the onedir versions shouldn’t have the problem https://docs.saltproject.io/salt/install-guide/en/latest/topics/upgrade-to-onedir.html#upgrade-to-onedir

I installed Salt from the APT repository [following the official install guide](https://docs.saltproject.io/salt/install-guide/en/latest/topics/install-by-operating-system/linux-deb.html).

Testing I can see that I am using Python 3.10 from the Salt onedir package with the below command.
I assumed this was connected to 3.12 because the issue only started after I upgraded my system.

```bash
$ salt-ssh --versions-report
Salt Version:
          Salt: 3006.9
 
Python Version:
        Python: 3.10.14 (main, Jun 26 2024, 11:44:37) [GCC 11.2.0]
```

PR #541, opened 2025-01-27
feat: add mysql user host

### LINKED ISSUE #540: Host '127.0.0.1' is not allowed to connect to this MySQL server
Hi @RobHooper, to finish configuring the digitalbuying app on ocp21, I need MySQL to accept connections from the Docker container.

Presently it doesn't seem to like tcp connections from localhost.

```
mysql -h 127.0.0.1 -P 3306 --protocol=tcp
mysql -h localhost -P 3306 --protocol=tcp
nc 127.0.0.1 3306
nc localhost 3306
```

Yet, the WordPress app can connect.

RobHooper:
Resolves #540

> [T]o finish configuring the digitalbuying app on ocp21, I need MySQL to accept connections from the Docker container.

Docker provisions a private IP range on the host for networking between containers (Using the `172.16.0.0/12` range by default). 
When the digitalbuying app connects to MySQL it appears as one of these private IP addresses, not localhost. 

You can view the digitalbuying network with the following command.
```
docker network inspect digitalbuying_default
```

This PR allows us to configure the MySQL user host in pillar.
I have allowed digitalbuying to connect over the entire private IP range as its IP range will change dynamically when networks are created and destroyed.

- [ ] Test the app works with these changes
- [ ] Document the `host` option
- [ ] Merge

### DIFF
--- a/pillar/digitalbuying.sls
+++ b/pillar/digitalbuying.sls
@@ -12,6 +12,7 @@ mysql:
   databases:
     digitalbuying:
       user: digitalbuying
+      host: "172.17.%"
 
 docker_apps:
   digitalbuying:
--- a/salt/mysql/init.sls
+++ b/salt/mysql/init.sls
@@ -61,6 +61,9 @@ remove test database:
 {% if 'password' in entry %}
     - password: "{{ entry.password }}"
 {% endif %}
+{% if 'host' in entry %}
+    - host: "{{ entry.host }}"
+{% endif %}
     - require:
       - service: mysql
 {% endfor %} {# users #}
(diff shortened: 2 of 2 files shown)


RobHooper:
@jpmckinney , I have already deployed this to cms and tested I can connect to MySQL from inside a docker container.

jpmckinney:
I think `docker compose run --rm --name my-app-shell cron python manage.py migrate --plan` but there might be an independent error interfering that I need to fix first.

jpmckinney (approved):

jpmckinney:
Okay, fixed that error and it looks like it connects :)

PR #542, opened 2025-02-06
feat: add postfix relay support

RobHooper:
This PR adds relay support to Postfix. We need this on portland-dev in order to send administrative email as they are being blocked by the ISP.

@jpmckinney , `pillar.smtp` is already configured on Credere, but I cannot see where it is in use in Salt?
I re-used `smtp` because it configures the same service this adds to Postfix.
However, if this clashes other parts of the code, we could rename the values in this PR to `pillar.postfix.relay`.

To Do:
- [ ] Create SMTP credentials and add to pillar private ([docs](https://ocdsdeploy.readthedocs.io/en/latest/deploy/services/aws.html#create-smtp-credentials))
- [ ] Merge and deploy to portland-dev

### DIFF
--- a/pillar/portland_dev.sls
+++ b/pillar/portland_dev.sls
@@ -1,7 +1,3 @@
-smtp:
-  host: email-smtp.us-west-2.amazonaws.com
-  port: 587
-
 network:
   host_id: ocp26
   ipv4: 20.106.239.92
--- a/pillar/portland_dev.sls
+++ b/pillar/portland_dev.sls
@@ -1,3 +1,7 @@
+smtp:
+  host: email-smtp.us-west-2.amazonaws.com
+  port: 587
+
 network:
   host_id: ocp26
   ipv4: 20.106.239.92
--- a/pillar/portland_dev.sls
+++ b/pillar/portland_dev.sls
@@ -1,3 +1,6 @@
+smtp:
+  relay: True
+
 network:
   host_id: ocp26
   ipv4: 20.106.239.92
--- a/salt/core/mail.sls
+++ b/salt/core/mail.sls
@@ -23,6 +23,38 @@ postfix:
       - pkg: postfix
     - onchanges:
       - debconf: postfix
+  file.keyvalue:
+    - name: /etc/postfix/main.cf
+    - key_values:
+        # On first run, myhostname may default to previous ISP value.
+        myhostname: "{{ pillar.network.host_id }}.{{ pillar.network.domain }}"
+{%- if 'smtp' in pillar %}
+        relayhost: "[{{ pillar.smtp.host }}]:{{ pillar.smtp.port }}"
+        smtp_sasl_auth_enable: "yes"
+        smtp_sasl_security_options: "noanonymous"
+        smtp_sasl_password_maps: "hash:/etc/postfix/sasl_passwd"
+        smtp_use_tls: "yes"
+        smtp_tls_note_starttls_offer: "yes"
+{%- endif %}
+    - separator: ' = '
+    - append_if_not_found: True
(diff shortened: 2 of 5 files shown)


jpmckinney:
> Create SMTP credentials and add to pillar private ([docs](https://ocdsdeploy.readthedocs.io/en/latest/deploy/services/aws.html#create-smtp-credentials))

I think we can reuse the credentials in `smtp.sls`?

pillar/portland_dev.sls
```diff
@@ -1,3 +1,7 @@
+smtp:
```
jpmckinney, line 1: Maybe just reuse the existing `smtp.sls` for portland-dev?

salt/core/mail.sls
```diff
@@ -23,6 +23,38 @@ postfix:
...
+    - name: /etc/postfix/main.cf
+    - key_values:
+        # On first run, myhostname may default to previous ISP value.
+        myhostname: "{{ pillar.network.host_id }}.{{ pillar.network.domain }}"
+{%- if 'smtp' in pillar %}
+        relayhost: "[{{ pillar.smtp.host }}]:{{ pillar.smtp.port }}"
+        smtp_sasl_auth_enable: "yes"
+        smtp_sasl_security_options: "noanonymous"
+        smtp_sasl_password_maps: "hash:/etc/postfix/sasl_passwd"
+        smtp_use_tls: "yes"
```
jpmckinney, line 36: We can either use `pillar.smtp.tls`, or change `conf-alertmanager.yml` to always set `require_tls: True`

salt/core/mail.sls
```diff
@@ -23,6 +23,38 @@ postfix:
       - pkg: postfix
     - onchanges:
       - debconf: postfix
+  file.keyvalue:
+    - name: /etc/postfix/main.cf
+    - key_values:
+        # On first run, myhostname may default to previous ISP value.
+        myhostname: "{{ pillar.network.host_id }}.{{ pillar.network.domain }}"
+{%- if 'smtp' in pillar %}
```
jpmckinney, line 31: We maybe want to check for a pillar.smtp.relay boolean or similar, otherwise prometheus will use the relay.

pillar/portland_dev.sls
jpmckinney: Or, I guess we can create a new sender in us-west, since it is closer.

RobHooper: Updated for `smtp.sls`

salt/core/mail.sls
RobHooper: Happy with this solution to prevent it's deployment on other servers.
I have added a `pillar.smtp.relay` boolean.

salt/core/mail.sls
RobHooper: I think we will and should always be using TLS.
I have updated this to hardcode alert manager `require_tls: True`.

salt/core/mail.sls
```diff
@@ -23,6 +23,38 @@ postfix:
       - pkg: postfix
     - onchanges:
       - debconf: postfix
+  file.keyvalue:
+    - name: /etc/postfix/main.cf
+    - key_values:
+        # On first run, myhostname may default to previous ISP value.
+        myhostname: "{{ pillar.network.host_id }}.{{ pillar.network.domain }}"
+{%- if 'relay' in pillar.smtp and pillar.smtp.relay == True %}
```
jpmckinney, line 31: Let's add `smtp` `relay: False` to common.sls, that way we can simplify this to `if pillar.smtp.relay`

RobHooper: Done.

PR #545, opened 2025-03-06
Sysctl redis

### LINKED ISSUE #524: Configure overcommit_ratio on PostgreSQL servers
unattended-upgrades on ocp22 has been failing reporting it is unable to allocate enough memory.
We have temporarily set `overcommit_memory` = `0` (the default value) to mitigate this.

PostgreSQL recommends `overcommit_memory` = `2` to help protect the service from accidental termination when memory resources are exhausted, however, this setting limits memory for all process causing issues particularly on smaller servers.

To resolve this we should increase `overcommit_ratio` from 50% to 90% / 100% allowing user space to access all available memory.

Further reading:
* https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-MEMORY-OVERCOMMIT
* https://www.kernel.org/doc/Documentation/vm/overcommit-accounting

RobHooper:
We are now configuring more kernel parameters with sysctl and for multiple services. This PR consolidates the setting of kernel parameters into one, dynamic core file.
This function needed extracting because `overcommit_memory` should be configured for both Redis and PostgreSQL.

@jpmckinney , I think, it would be beneficial to add a prompt in the docs to configure `overcommit_memory` when Redis containers in Docker are added. I am not sure where to add this, or how big of a section it should be, please can you advise?

Once merged, the following needs actioning:
- [ ] Remove `/etc/sysctl.d/99-swappiness.conf`

Closes #524

### DIFF
--- a/docs/develop/update/docker.rst
+++ b/docs/develop/update/docker.rst
@@ -113,6 +113,16 @@ Reference:
 -  `Compose specification <https://docs.docker.com/compose/compose-file/>`__
 -  `Use Compose in production <https://docs.docker.com/compose/production/>`__
 
+Redis
+~~~~~
+
+If the Docker Compose file describes a Redis service, add, in the server's Pillar file:
+
+.. code-block:: yaml
+
+   vm:
+     overcommit_memory: 1
+
 Configure Docker app
 --------------------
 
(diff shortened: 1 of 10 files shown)


jpmckinney (commented):
Looks good. Had one comment and I made one commit.

jpmckinney (approved):

PR #548, opened 2025-04-02
feat: unifying backup date format

RobHooper:
The `tar` command interprets colons in filenames as a remote server to connect too.
This quirk of tar meant that working with our PostgreSQL backups was not straightforward, and importantly, could delay recovery in critical moments.

This PR changes backup file date formats to only include a basic character set while still following iso 8601 standardisation.

I have also added the time to file backups allowing us to backup multiple times in one day.


jpmckinney (approved):
Looks good! I guess we'll just let the files with the old name format expire on S3?

RobHooper:
> Looks good! I guess we'll just let the files with the old name format expire on S3?

Agreed, the old named files will rotate out over the next month.

PR #566, opened 2025-07-11
Cove US re-deploy and TinyProxy

### LINKED ISSUE #547: Re-deploy a London Linode to a US region
And add tinyproxy, if not already in use.

I think the CMS server might be the simplest to redeploy. It's low traffic and has few files to transfer.

* docs: Requires migrating a lot of files. High traffic.
* cove: Requires migrating *some* files. Moderate traffic.
* prometheus: Requires redeploying most servers to reconfigure their firewall.

RobHooper:
Closes #547

### DIFF
--- a/pillar/cove.sls
+++ b/pillar/cove.sls
@@ -11,16 +11,22 @@ x-env: &env
   REQUESTS_CACHE_EXPIRE_AFTER: 0 # EXPIRE_IMMEDIATELY
 
 network:
-  host_id: ocp18
-  ipv4: 176.58.107.239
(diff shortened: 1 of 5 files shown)


RobHooper:
@jpmckinney , This PR reuses the existing Tiny Proxy pillar file, I just wanted to double check that is OK.
This file has a `url` variable which is hard coded for ocp19, however, I cannot see this actually being used anywhere. Can I remove the `url` variable?

https://github.com/open-contracting/deploy/blob/main/pillar/tinyproxy.sls#L8

PR #580, opened 2025-08-28
feat: reduce nr_hugepages

### LINKED ISSUE #517: Update nr_hugepages on ocp13 and ocp23
I reduced shared_buffers to 8GB, because apparently there might not be much advantage to go over that: https://github.com/jberkus/annotated.conf/blob/master/postgresql.10.simple.conf#L84-L87

I originally set nr_hugepages following this process:
https://www.postgresql.org/docs/14/kernel-resources.html#LINUX-HUGE-PAGES

As documented here: https://ocdsdeploy.readthedocs.io/en/latest/develop/update/postgres.html

However, this requires us to "start PostgreSQL without huge pages enabled"

I think there is no great consequence to having a larger nr_hugepages than necessary, but logging here in case easy to do.

RobHooper:
Closes #517

### DIFF
--- a/pillar/kingfisher_main.sls
+++ b/pillar/kingfisher_main.sls
@@ -51,7 +51,6 @@ users:
 
 vm:
   # https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-HUGE-PAGES
-  # https://github.com/open-contracting/deploy/issues/517
   nr_hugepages: 4300
 
 ntp:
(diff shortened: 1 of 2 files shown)


RobHooper:
After deploying we need to restart PostgreSQL.

To check the new setting has taken effect:
```
grep HugePages /proc/meminfo
```

jpmckinney (approved):

PR #593, opened 2025-12-05
feat: fix logging behind cloudflare proxy

RobHooper:


jpmckinney (approved):
Looks good. I made a couple commits (write to conf-available, then enable).

PR #623, opened 2026-05-12
refactor sync-to-s3.sh per directory

### LINKED ISSUE #618: S3 sync: Ignore requests.queue directories
Scrapyd has directories like:

```
/home/collect/scrapyd/jobs/kingfisher/germany_digiwhist/8c630d3ab5ec11f088609c6b004cdfa7/requests.queue
```

These contain `active.json` which is just `[]` for finished jobs.

Since these change frequently for running jobs, they cause an error when syncing:

> An error occurred (IncompleteBody) when calling the PutObject operation: You did not provide the number of bytes specified by the Content-Length HTTP header

We should ignore these directories.

RobHooper:
This PR modifies the sync-to-s3.sh to sync one directory rather than many, this allows us to add directory specific features (exclude paths #618) without the feature being applied to all paths.
The main complexity of backing up directories has been moved into Salt creating a cron for each sync path and setting features as required - this should prove more extendable in the future.

Closes #618

Connected tasks:
- [ ]  Remove SYNC_DIRECTORIES from settings.local

### DIFF
--- a/salt/aws/files/sync-to-s3.sh
+++ b/salt/aws/files/sync-to-s3.sh
@@ -29,36 +29,17 @@ if [ -z "$S3_SYNC_BUCKET" ]; then
     exit 4
 fi
 
-AWS_ARGS=(--only-show-errors --delete)
-
-# Parse script arguments
-while [[ $# -gt 0 ]]; do
-    case $1 in
-    --exclude)
-        if [ -z "$2" ]; then
-            echo "Missing path."
-            exit 1
-        fi
-        AWS_ARGS+=(--exclude "\"$2\"")
-        shift 2
-        ;;
-    -*)
-        echo "Unknown option"
-        exit 1
-        ;;
-    *)
-        DIRECTORY=$1
-        SAFENAME=${DIRECTORY/#\//}
-        SAFENAME="${SAFENAME/%\//}"
-        shift
-        ;;
-    esac
-done
-
-if [ -e "$DIRECTORY" ]; then
+if [ ! -e "$1" ]; then
     echo "Error: DIRECTORY isn't set or doesn't exist."
     exit 5
 fi
+
+DIRECTORY=$1
+SAFENAME=${DIRECTORY/#\//}
+SAFENAME="${SAFENAME/%\//}"
+shift
+
+AWS_ARGS=(--only-show-errors --delete "${@}")
 
 set +e
 $AWS_CLI s3 sync "${AWS_ARGS[@]}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
--- a/salt/aws/files/sync-to-s3.sh
+++ b/salt/aws/files/sync-to-s3.sh
@@ -42,5 +42,5 @@ shift
 AWS_ARGS=(--only-show-errors --delete "${@}")
 
 set +e
-$AWS_CLI s3 sync "${AWS_ARGS[@]}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
+$AWS_CLI s3 sync --only-show-errors --delete "${@}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
 set -e
--- a/salt/aws/files/sync-to-s3.sh
+++ b/salt/aws/files/sync-to-s3.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Backup multiple directories and upload to AWS S3
+# Sync a single directory into S3.
 
 set -euo pipefail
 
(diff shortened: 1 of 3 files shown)


salt/aws/files/sync-to-s3.sh
```diff
@@ -24,16 +24,42 @@ if [ ! -x "$AWS_CLI" ]; then
...
 done
+
+if [ -e "$DIRECTORY" ]; then
+    echo "Error: DIRECTORY isn't set or doesn't exist."
+    exit 5
+fi
+
+set +e
+$AWS_CLI s3 sync "${AWS_ARGS[@]}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
+set -e
```
RobHooper, line 44: This final `set -e` does nothing, I intend for it to prevent human error keeping the flag active if the script was extended in the future.

salt/aws/sync.sls
```diff
@@ -16,16 +16,10 @@ include:
   file.managed:
     - contents: |
         MAILTO=root
-        15 03,15 * * * root /home/sysadmin-tools/bin/sync-to-s3.sh
+{%- for directory, entry in pillar.sync.directories|items %}
+{%- set minute = (loop.index0 * 5) % 60 %}
+        {{minute}} 03,15 * * * root /home/sysadmin-tools/bin/sync-to-s3.sh {{ directory }}
+        {%- if entry and 'exclude' in entry %} --exclude "{{ entry.exclude }}"{% endif %}
```
jpmckinney: Just checking: This puts the cron job onto two lines. Does cron read onto the second line?

salt/aws/files/sync-to-s3.sh
```diff
@@ -24,16 +24,42 @@ if [ ! -x "$AWS_CLI" ]; then
...
     exit 4
 fi
 
-for DIRECTORY in "${SYNC_DIRECTORIES[@]}"; do
-    SAFENAME="${DIRECTORY/#\//}"
-    SAFENAME="${SAFENAME/%\//}"
+AWS_ARGS=(--only-show-errors --delete)
 
-    set +e
-    $AWS_CLI s3 sync "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/" --only-show-errors --delete 2>&1 | grep -v "You did not provide the number of bytes specified by the Content-Length HTTP header"
```
jpmckinney, line 37: Do we not need this anymore?

` 2>&1 | grep -v "You did not provide the number of bytes specified by the Content-Length HTTP header"`

salt/aws/sync.sls
jpmckinney: Also, maybe do (I think `items` is safe on None?):

{%- for option, value in entry|items %} --{{ option }} "{{ value }}"{% endfor %}

salt/aws/files/sync-to-s3.sh
```diff
@@ -24,16 +24,42 @@ if [ ! -x "$AWS_CLI" ]; then
...
+            exit 1
+        fi
+        AWS_ARGS+=(--exclude "\"$2\"")
+        shift 2
+        ;;
+    -*)
+        echo "Unknown option"
+        exit 1
+        ;;
+    *)
```
jpmckinney: This seems a bit paranoid, as we control the Pillar data, and should only be setting real args. I think we can pass through args directly.

salt/aws/sync.sls
RobHooper: > This puts the cron job onto two lines. Does cron read onto the second line?

The cron stays on one line thanks to `{%-`. I put the code across two lines for readability.

> {%- for option, value in entry|items %} --{{ option }} "{{ value }}"{% endfor %}

This is much more dynamic - implemented.
Salt does not like when `entry` is None but setting a default value (`{}`) in the loop fixes this.

salt/aws/files/sync-to-s3.sh
RobHooper: I am paranoid of word splitting which is how I ended up with this solution, however, I have found and committed a new solution that handles this while also passing args directly.
The one assumption the new code now makes is that the backup directory will be the first arg, we need to capture and format the directory before passing it through to make the S3 paths.

salt/aws/files/sync-to-s3.sh
RobHooper: We may still want to ignore these alerts (files changing as the script runs).
I want to see how much the exclude functionality reduces these warnings first.

salt/aws/files/sync-to-s3.sh
```diff
@@ -24,16 +24,23 @@ if [ ! -x "$AWS_CLI" ]; then
...
+shift
+
+AWS_ARGS=(--only-show-errors --delete "${@}")
 
-    set +e
-    $AWS_CLI s3 sync "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/" --only-show-errors --delete 2>&1 | grep -v "You did not provide the number of bytes specified by the Content-Length HTTP header"
-    set -e
-done
+set +e
+$AWS_CLI s3 sync "${AWS_ARGS[@]}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
```
jpmckinney: I suppose we can remove the `AWS_ARGS` assignment and do:


```suggestion
$AWS_CLI s3 sync --only-show-errors --delete "${@}" "$DIRECTORY" "s3://$S3_SYNC_BUCKET/$SAFENAME/"
```

salt/aws/sync.sls
```diff
@@ -16,16 +16,10 @@ include:
   file.managed:
     - contents: |
         MAILTO=root
-        15 03,15 * * * root /home/sysadmin-tools/bin/sync-to-s3.sh
+{%- for directory, entry in pillar.sync.directories|items %}
+{%- set minute = (loop.index0 * 5) % 60 %}
+        {{minute}} 03,15 * * * root /home/sysadmin-tools/bin/sync-to-s3.sh {{ directory }}
+        {%- for option, value in (entry or {}) | items %} --{{ option }} "{{ value }}"{% endfor %}
```
jpmckinney: nit: I think our [style guide](https://ocdsdeploy.readthedocs.io/en/latest/develop/style/language.html) prefers `default`, though we could change that for the case of empty dict and empty list (only affects 7 locations that do `default([])`). The `or` version is briefer. This change can be done in a separate cleanup commit after the PR.

```suggestion
        {%- for option, value in entry | default({}) | items %} --{{ option }} "{{ value }}"{% endfor %}
```

jpmckinney (approved):

salt/aws/sync.sls
RobHooper: I prefer `default` for readability however the above code breaks in this situation leading me to the `or` solution.

Digging further into this I have found that we need `default({}, true)` instead.
The problem was that entry has the value of None but `default` by default is only changes "undefined".

https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.default

PR #635, opened 2026-07-08
docs: downtime protocol and troubleshooting

### LINKED ISSUE #569: Update downtime protocol
https://ocdsdeploy.readthedocs.io/en/latest/reference/downtime.html

We can add details about:

- alerts (uptime monitors)
- common diagnostic steps
- common solutions (e.g. contact the hosting provider)
- link to relevant pages e.g. if the server is a total loss, we need to [re-deploy](https://ocdsdeploy.readthedocs.io/en/latest/deploy/create_server.html) the server and restore from backups (perhaps @yolile  can review the [Testing backups](https://ocdsdeploy.readthedocs.io/en/latest/maintain/backup.html) instructions, to see if they are clear enough to use as instructions for restoring a server)

At the same time, we can shorten or clarify the current text as needed.

RobHooper:
Resolves #569

### DIFF
--- a/docs/reference/downtime_troubleshooting.rst
+++ b/docs/reference/downtime_troubleshooting.rst
@@ -33,7 +33,8 @@ Get single service information, for example apache2:
 System logs are accessible via journalctl -u example.service
 They are also directly readable in /var/log
 
-Notably: 
+Notably:
+
 - All system alerts: /var/log/syslog
 - Apache website: /var/log/apache2
 - NGINX site logs: /var/log/nginx
@@ -102,7 +103,7 @@ High Disk Space
 View total disk space
 
 .. code-block:: bash
-   
+
    df -h
 
 View disk usage of a directory and its contents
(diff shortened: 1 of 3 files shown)


docs/reference/downtime_troubleshooting.rst
```diff
@@ -0,0 +1,336 @@
...
+Restart application, for example cove-ocds:
+
+.. code-block:: bash
+
+   sudo -u deployer docker compose -f /data/deploy/cove-ocds/docker-compose.yaml pull
+   sudo -u deployer docker compose -f /data/deploy/cove-ocds/docker-compose.yaml down
+   sudo -u deployer docker compose -f /data/deploy/cove-ocds/docker-compose.yaml up -d
+
+
+When running `docker compose up -d` by Docker automatically updates and pulls the latest container images, however this significantly increases downtime so we prefer to pull images in advance.
```
jpmckinney, line 227: Hmm, if there's an outdated image, `up` will use it, unless you set `--pull always` or `pull_policy: always`

docs/reference/downtime_troubleshooting.rst
```diff
@@ -0,0 +1,336 @@
...
+- Are you blocked by the firewall?
+  - Is there an external firewall blocking you?
+  - Have you port knocked? (on non-docker servers)
+
+If the server is inaccessible and completely offline you may need to jump to :ref:`restarting the server via the ISP<isp-reboot>`.
+
+Linode Physical Host Outage
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We have seen the occasional issue with the underlying physical host running our VPS taking fault, this has then taken the OCP server(s) offline and prevent the Linode admin interface from responding. If the issue is limited to only one host this will not be posted on the public Linode Status Page, instead we will be contacted directly on a support ticket. This support ticket will set expectations of how Linode will be responding to the issue and timescales for a resolution.
```
jpmckinney, line 200: Re: admin interface not responding, what does that mean specifically: like power off and reboot don't work?

docs/reference/downtime_troubleshooting.rst
```diff
@@ -0,0 +1,336 @@
...
+#. "Quick Actions", toggle "Under Attack Mode"
+
+Total Server Loss
+-----------------
+
+In the rare circumstance of a total server loss the only option is to rebuild a new server and recover data from backups.
+
+#. Create new server
+   #. https://ocdsdeploy.readthedocs.io/en/latest/deploy/create_server.html
+   #. Either use the latest supported operating system image or the version used on the previous server. Using the latest image may require additional configuration changes.
```
jpmckinney, line 329: I think we'll always use the previous server's version, unless we happen to be very close to a planned upgrade.

docs/reference/downtime_troubleshooting.rst
RobHooper: By default when the "latest" tag is used `up` will always pull new images.

I have been caught out a few times by this resulting in a longer outage than necessary, so it has become part of my Docker run book.

https://docs.docker.com/reference/compose-file/services/#pull_policy

docs/reference/downtime_troubleshooting.rst
RobHooper: I wrote this just days after responding to the last Linode outage. I believe the interface allowed me to issue reboot and shutdown commands but nothing happened, the server continued reporting as running. 
As opposed to the interface itself freezing or not responding.

jpmckinney: Aha. I added this back now:

> A symptom of a physical host outage is a server not responding to *Power Off* or *Reboot* actions.

docs/reference/downtime_troubleshooting.rst
jpmckinney: Aha, I've restored with that link.

PR #636, opened 2026-07-13
feat: restart scrapyd on failure

### LINKED ISSUE #563: Scrapyd failing to start after an unclean shutdown
After an unclean shutdown Scrapyd fails to start during server boot up, this leaves the service offline until it is manually started.
We have seen this issue on Registry multiple times.

One solution to help start Scrapyd on boot would be for systemd to retry the service, for example:

```
[Unit]
StartLimitInterval=200
StartLimitBurst=5

[Service]
Restart=always
RestartSec=30
```

RobHooper:
Resolves #563

I am assuming that Scrapyd returns a non-zero exit code on failure. If this is not the case we can use `Restart=always`
[Further reading](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html?__goaway_challenge=meta-refresh&__goaway_id=5b1df7a2e2758397ae131c033abe1383#Restart=).

### DIFF
--- a/salt/core/systemd/files/scrapyd.service
+++ b/salt/core/systemd/files/scrapyd.service
@@ -1,6 +1,9 @@
 [Unit]
 Description=Scrapyd
 After=network.target
+# https://github.com/open-contracting/deploy/issues/563
+StartLimitInterval=200
+StartLimitBurst=5
 
 [Service]
 User={{ user }}
@@ -18,6 +21,9 @@ WorkingDirectory={{ entry.appdir }}
 ExecStart={{ entry.appdir }}/.ve/bin/scrapyd --logfile=/var/log/scrapyd/scrapyd.log
 # https://github.com/open-contracting/kingfisher-collect/issues/606#issuecomment-953896571
 LimitNOFILE=512000
+# https://github.com/open-contracting/deploy/issues/563
+Restart=on-failure
+RestartSec=30
 
 [Install]
 WantedBy=multi-user.target
--- a/salt/core/systemd/files/scrapyd.service
+++ b/salt/core/systemd/files/scrapyd.service
@@ -1,8 +1,7 @@
 [Unit]
 Description=Scrapyd
 After=network.target
-# https://github.com/open-contracting/deploy/issues/563
-StartLimitInterval=200
+StartLimitIntervalSec=300
 StartLimitBurst=5
 
 [Service]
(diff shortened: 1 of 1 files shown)


jpmckinney:
Yes, it has non-zero exit on failure (e.g. when running two processes on the same port).

salt/core/systemd/files/scrapyd.service
```diff
@@ -1,6 +1,9 @@
 [Unit]
 Description=Scrapyd
 After=network.target
+# https://github.com/open-contracting/deploy/issues/563
+StartLimitInterval=200
```
jpmckinney: I think this is `StartLimitIntervalSec` now?

So this would mean it tries to restart at most 5 times every 3.333 minutes? I figure we can bump to 300 for the pleasure of round numbers.

salt/core/systemd/files/scrapyd.service
```diff
@@ -1,6 +1,9 @@
 [Unit]
 Description=Scrapyd
 After=network.target
+# https://github.com/open-contracting/deploy/issues/563
```
jpmckinney: We can omit the issue comments. The provenance will be clear from git/PR history.

salt/core/systemd/files/scrapyd.service
```diff
@@ -18,6 +21,9 @@ WorkingDirectory={{ entry.appdir }}
 ExecStart={{ entry.appdir }}/.ve/bin/scrapyd --logfile=/var/log/scrapyd/scrapyd.log
 # https://github.com/open-contracting/kingfisher-collect/issues/606#issuecomment-953896571
 LimitNOFILE=512000
+# https://github.com/open-contracting/deploy/issues/563
+Restart=on-failure
```
jpmckinney, line 21: Let's put LimitNOFILE last, e.g. this identity runs, with these privileges and in this context, this specific command, which it can retry, and once it's running here are the process' resource limits.

salt/core/systemd/files/scrapyd.service
RobHooper: Both fixed.

salt/core/systemd/files/scrapyd.service
RobHooper: Fixed

salt/core/systemd/files/scrapyd.service
RobHooper: Fixed.

jpmckinney (approved):

PR #640, opened 2026-08-04
Add Ubuntu 26.04

RobHooper:
Also updating Linode server build documentation, updating default versions and removing legacy configuration.


salt/core/ntp/init.sls
```diff
@@ -1,21 +1,40 @@
...
+{% if grains.osmajorrelease|int >= 26 %}
+chrony:
+  service.running:
+    - name: chrony
+
+chrony-reload:
+  cmd.wait:
+    - name: chronyc reload sources
+
+/etc/chrony/sources.d/ntp-pools.sources:
```
RobHooper, line 11: @jpmckinney , with Chrony we can now use NTS (Network Time Security) replacing NTP.
The current PR code re-implements our previous NTP setup disabling the defaults but we could take this in a different direction. The main restriction on NTS is the lack of sources to pick between [1].

If we move to NTS the best sources in my opinion are Ubuntu and Cloudflare, however these are all location-less potentially increasing latency when fetching updates.


[1] This is the best list of NTS servers I have found so far:
https://github.com/jauderho/nts-servers#the-list

jpmckinney: I think either of those sources is fine

RobHooper: Updated

salt/core/apt/init.sls
```diff
@@ -12,7 +12,6 @@ needrestart:
   pkg.removed:
     - name: needrestart
 
-# https://www.phusionpassenger.com/library/install/apache/install/oss/bionic/
 # gnupg depends on dirmngr. gnupg2 is a dummy package for gnupg.
```
jpmckinney: We can remove this line too - I think it was about reconciling the link’s instructions with our package list

pillar/common.sls
```diff
@@ -23,14 +23,29 @@ ssh:
...
+nts:
+  time.cloudflare.com:
+  1.ntp.ubuntu.com:
+  2.ntp.ubuntu.com:
+  3.ntp.ubuntu.com:
+  4.ntp.ubuntu.com:
+  ntp-bootstrap.ubuntu.com:
+    context: iburst maxsources 1 nts certset 1
+
+
```
jpmckinney: Remove consecutive newline

salt/apache/files/conf/letsencrypt.conf
```diff
@@ -2,17 +2,31 @@
     MDContactEmail sysadmin@open-contracting.org
     MDCertificateAgreement accepted
     MDCAChallenges http-01
+{% if grains.osmajorrelease|int >= 26 %}
+    # Store mod_md data outside /etc due to systemd filesystem sandboxing.
```
jpmckinney: Use {#- #} to avoid writing this comment on the server

salt/apache/files/conf/letsencrypt.conf
```diff
@@ -2,17 +2,31 @@
...
+    MDStoreDir /var/lib/apache2/md
+{% endif %}
```
jpmckinney:
```suggestion
{%- endif %}
```

salt/apache/files/conf/letsencrypt.conf
```diff
@@ -2,17 +2,31 @@
     MDContactEmail sysadmin@open-contracting.org
     MDCertificateAgreement accepted
     MDCAChallenges http-01
+{% if grains.osmajorrelease|int >= 26 %}
```
jpmckinney:
```suggestion
{%- if grains.osmajorrelease|int >= 26 %}
```

salt/core/firewall/init.sls
```diff
@@ -38,6 +38,9 @@ save iptables rules:
     - onchanges:
       - file: /home/sysadmin-tools/firewall-settings.local
       - file: /home/sysadmin-tools/bin/firewall.sh
+    - success_retcodes:
```
jpmckinney, line 38: Move above onchanges (we put all hooks last)

salt/core/ntp/init.sls
```diff
@@ -1,21 +1,40 @@
...
+    - watch_in:
+      - cmd: chrony-reload
+
+/etc/chrony/sources.d/ubuntu-ntp-pools.sources:
+  file.comment:
+    - regex: "^pool "
+    - backup: False
+    - watch_in:
+      - cmd: chrony-reload
+
```
jpmckinney: Remove extra newline

salt/core/sshd/files/customization.conf
```diff
@@ -1,3 +1,8 @@
+# We set both PermitRootLogin and PasswordAuthentication for two reasons:
```
jpmckinney: Use a Jinja comment with %- and -% at both ends

salt/core/mail.sls
```diff
@@ -33,7 +33,11 @@ postfix:
         smtp_sasl_auth_enable: "yes"
         smtp_sasl_security_options: "noanonymous"
         smtp_sasl_password_maps: "hash:/etc/postfix/sasl_passwd"
+{% if grains.osmajorrelease|int >= 26 %}
+        #  Depreciated in Postfix 3.9, instead specify 'smtp_tls_security_level'
```
jpmckinney: Deprecated typo. I’m unclear on the logic, we only set this in Ubuntu 26, but isn’t that where it’s deprecated?

docs/deploy/create_server.rst
```diff
@@ -34,6 +34,10 @@ Create the server via the :ref:`host<hosting>`'s interface.
          #. Set *Linode Label* to the server's FQDN (e.g. ``ocp99.open-contracting.org``)
          #. Set *Add Tags* to either *Production* or *Development*
          #. Set *Root Password* to a `strong password <https://www.lastpass.com/features/password-generator>`__, and save it to OCP's `LastPass <https://www.lastpass.com>`__ account
+         #. Set *Public Interface Firewall* to *No firewall*, unless the server uses Docker.
```
jpmckinney: Let’s collapse to one bullet:

```
If using Docker, set *Public Interface Firewall* *Create Firewall* and :ref:`configure an external firewall starting at step 3<docker-firewall>`. Otherwise, set to *No firewall*.

RobHooper: Updated

pillar/common.sls
RobHooper: Fixed

salt/core/mail.sls
RobHooper: Typo fixed.
Logic corrected - this variable is only set on systems older than 26.

salt/core/sshd/files/customization.conf
RobHooper: Comment syntax changed.
This file is now marked with `template: jinja` so the comment is not set on servers.

salt/core/mail.sls
jpmckinney: Ok, and just checking smtp_tls_security_level is supported on old versions?

RobHooper: Yes `smtp_tls_security_level` has been supported for a while now and is already set on our current servers.
The deprecated setting `smtp_use_tls` is now inferred rather than a stand alone variable.

jpmckinney: Aha, so is smtp_use_tls not inferred in old versions and that's why we still need it?

RobHooper: Re-reading the Postfix documentation I have decided to remove `smtp_use_tls` completely.
The new `smtp_tls_security_level` option is already set on all servers so `smtp_use_tls` hasn't been actually making any difference anyway.
https://www.postfix.org/postconf.5.html#smtp_tls_security_level

